Issue #1, opened 2026-07-01
PRD: Distributed Agent Research System


thecolab-clawd:
**Version 1.0 | 2026-06-30 | Ready for Implementation**

## TL;DR
- GitHub Issues = task queue
- `work.sh` claims a task, runs Claude agent with standardized method (`CONTRIBUTING.md`), creates PR
- `review.sh` adversarially verifies findings before merge
- **Result:** Transparent, scalable, zero duplication, full audit trail

## System Flow
```
Issue -> work.sh (claim + research) -> PR -> review.sh (adversarial check) -> PASS: Merge | NEEDS_WORK: Revise
```

## Key Scripts

**work.sh** - Grab unclaimed issue -> mark in-progress -> pass to Claude (task + method) -> agent researches -> create PR -> label review-needed

**review.sh** - Grab PR needing review -> pass to adversarial agent ("refute this") -> if PASS: merge + done; if NEEDS_WORK: request changes + revert to in-progress

**CONTRIBUTING.md** - Research method all agents follow:
1. Clarify the question
2. Choose approach (Workflow for multi-agent, Agent for single)
3. Research & cite (every claim >=1 source)
4. Verify surprises (>=2 sources)
5. Mark confidence (High/Medium/Low)
6. Format findings (use PR template)

## Repo Structure
```
├── README.md
├── CONTRIBUTING.md
├── work.sh
├── review.sh
├── .github/ISSUE_TEMPLATE/research-task.md
├── .github/PULL_REQUEST_TEMPLATE/research-finding.md
└── research/findings/ (merged findings live here)
```

## GitHub Labels
- Status: unclaimed | in-progress | review-needed | done
- Domain: child-welfare | grant-access | civic-transparency | ai-policy | biosecurity
- Task: research/task | research/finding

## Phases
- **Phase 1 (Week 1):** Repo + scripts + templates + end-to-end test on 2 tasks
- **Phase 2 (Week 2):** GitHub Actions automation
- **Phase 3 (Week 3):** Metrics + dashboard
- **Phase 4 (Week 4+):** Scale

## Success Story Example
Monday 9am: Issue #42 created ("Starship impact on child welfare?")
10am: Researcher claims + runs work.sh
10-12:30pm: 3 agents research in parallel
12:30pm: PR #88 created, needs review
1:30pm: review.sh finds 2 weak points
2-3pm: Researcher responds with NZ data
3:30pm: review.sh PASS -> merges
[...]

Issue #5, opened 2026-07-01
[Research] How many grant funders operate in NZ, and where does eligibility data live?


adam91holt:
**The question**
How many grant funders operate in New Zealand (philanthropic trusts, gaming/community trusts, councils, government), and in what form is their eligibility/criteria data available — machine-readable, PDF, web pages, or not published at all?

**Why it matters / what decision it informs**
Whether a grant-matching tool for small charities is feasible depends entirely on whether eligibility data can be assembled. This scopes that.

**Part of** the grant-access Discover issue.

**Output:** a cited finding in `research/findings/grant-access/` using `research/TEMPLATE.md`. Follow the method — cite everything, mark confidence, say what you couldn't verify.

_Good first issue: bounded, mostly desk research, high value._

Part of #2. Stream: #2.

mcinteerj:
Claimed and completed — worked by Jake's Claude agent. Finding is up as PR #11 (`research/findings/grant-access/nz-grant-funder-count-and-eligibility-data.md`), awaiting adversarial review. Note: I don't have label permissions on this repo, so a maintainer may want to flip `status: available` → `status: in-review`.

adam91holt:
🔁 Adversarial review of PR #11 requested changes — flipping to **status: changes-requested** and assigning @mcinteerj so their next `start_work.sh` loop picks the rework up (new pushback flow from #33).

adam91holt:
Reopening: this was closed as completed, but the finding never actually landed — it was accidentally merged via #29 and reverted in #31, and PR #11 is still open with a changes-requested adversarial review. #21 covers a narrower follow-up (the Charities Register route), not this broader question. Putting it back in the rework queue so @mcinteerj's next `start_work.sh` loop addresses the review feedback on PR #11.

mcinteerj:
Rework status: PR #11 round-3 fixes are pushed (`692bc7f`, after the pushback label was set) with a point-by-point reply on the PR — ready for re-review. `start_work.sh`'s rework path can't run from this account (fork PR branches aren't on origin + no triage for label edits — details in https://github.com/thecolab-ai/the-for-good-project/issues/20#issuecomment-4861849529), so this issue just needs a maintainer to flip `changes-requested → in-review`.

Issue #2, opened 2026-07-01
[Discover] Small NZ charities miss grants they're eligible for


adam91holt:
**The problem**
New Zealand has hundreds of grant funders (philanthropic trusts, gaming trusts, councils, government), but discovery is manual and fragmented. Small charities and community groups — often run by volunteers — miss funding they're eligible for simply because finding and matching to the right grants is hard and time-consuming.

**Why it matters**
The unclaimed/under-applied pool is plausibly significant, and the groups worst affected are the smallest ones doing frontline work. _(Figures to be established — see the research questions below.)_

**Questions this raises** (candidates for Research issues)
- How many grant funders operate in NZ, and where does their eligibility data actually live?
- What do small charities currently use to find grants, and where does it break down?
- Is there a machine-readable source of NZ grants, or would one need to be assembled?

**Next step:** open Research issues for the questions above. See CONTRIBUTING.md.

karenjleacock:
Check out https://www.fundsorter.com/ 
It's about a year old? They are also doing great things in the training space.
I can't answer the questions you pose (and don't have spare tokens), but it feels like they are leading in this space in NZ.

adam91holt:
🤖 @adam91holt is starting work on this via `start_work.sh` (agent: `codex`).

adam91holt:
Discovery framing complete for Stream: #2.

I opened four research child issues:
- #45 — Map NZ grant funders and eligibility-source locations
- #46 — How small NZ charities find grants and where the workflow fails
- #47 — Audit machine-readable NZ grants data and schema gaps
- #48 — Measure whether eligible small charities under-apply for grants

I also drafted a cited framing note for the PR. The key limit: the scale of charities and source fragmentation is supported, but the claim that eligible small charities miss grants remains a hypothesis until the child research issues test it.

adam91holt:
🔍 Work submitted in #49 — moving to **in review**. It needs an adversarial review from a *different identity than the author* before it can merge (see `review_work.sh`).

adam91holt:
🧩 Synthesis draft ready in #105 — a human **steward** now reviews the rollup, fixes any overreach, writes the direction decision (go deeper / pivot / proceed / park), and merges. This is gate **G1** (docs/STREAMS.md); the stream stays parked at **awaiting-direction** until then.

adam91holt:
[...]

PR #14, opened 2026-07-01
feat(web): problem-chains view on the board

### LINKED ISSUE #41: build: problem-chains board view, grouped by stream labels
Part of #30. Split from #37.

**Goal:** a third board view showing each stream as a nested "problem chain" — child issues under their parents, PR chips (open/merged), status badges — so the Discover → Research → Ideate → Build pipeline is visible per problem, not just as stage columns.

**Head start — salvage, don't start from scratch:** closed PR #14 built most of the rendering (`ChainTree` component, `lineage.ts`, filter dimming; screenshot on the PR). Fetch it with `git fetch origin pull/14/head` and lift what's useful. It predates the stream model, so the foundations change:

1. **Group chains by the `stream:` label** — the canonical, server-maintained grouping (validated + propagated by `stream-sync.yml`). Body-ref parsing may still shape the tree *inside* a stream, but must be **line-anchored** (`^Part of #n`) — unanchored prose matching mis-parents issues.
2. **Fix the cycle crash found in PR #14's adversarial review** — promoted cycle roots keep child pointers, so `sortChildren` recurses forever (`RangeError`) and takes down the whole Board page.
3. **Build against current main** — streams are live; render the new statuses (`changes-requested`, `needs-synthesis`, `awaiting-direction`) and remember discover roots now stay open for a stream's life (#40).

Coordinate with #37 (per-stream pages) — the chain tree is a natural component of those pages.

ian-patel:
## What this is

**Stage:** 🔨 Build (site infrastructure — `web/`)
**Domain:** n/a (dashboard)

## Summary

Adds a third view to the board — **problem chains** — that shows each problem's full lineage: child issues nest under their parent via the "Part of #n" body convention, and pull requests attach to the issues they close (via "Closes/Fixes/Resolves #n"). This makes the Discover → Research → Ideate → Build pipeline the README describes actually visible per problem, instead of only as stage columns.

Details:
- New `web/src/lib/lineage.ts` builds the trees client-side from the existing snapshot (no build-script or snapshot-format changes needed).
- New `ChainTree` component renders a card per root problem with indented children, PR chips (open/merged), status badges, and assignee avatars.
- Filters still work in chains view: a chain is shown if any node matches; non-matching nodes render dimmed so the chain keeps its shape. Closed issues stay visible as completed steps (struck through, "done").
- Defensive handling: unknown "Part of" targets and cycles fall back to root, so every issue always renders.

## Method checklist

- [x] No personal or identifying data; no overstated or fabricated claims
- [x] Files are in the right place and follow the relevant template
- (Findings-specific items n/a — this is a site change.)

## For the reviewer

- Verified locally: `npm ci && npm run lint && npm run build` pass; tested in the browser with real snapshot data (PR chips attach to #5/#6) and with a synthetic "Part of #2" chain to confirm nesting, filter dimming, and closed-issue rendering.
- The weakest point to push on: parent detection is regex on the issue body ("part of #n", first match, case-insensitive). If the project later adopts GitHub sub-issues or a different linking convention, `lineage.ts` is the one place to change.

Screenshot 2026-07-02 at 11 35 01 AM

Closes #41.

### DIFF
--- a/web/src/components/shared/ChainTree.tsx
+++ b/web/src/components/shared/ChainTree.tsx
@@ -1,0 +1,85 @@
+import { Link } from "react-router-dom";
+import { GitMerge, GitPullRequest, CheckCircle2 } from "lucide-react";
+import type { ChainNode } from "@/lib/lineage";
+import type { IssueLite } from "@/lib/types";
+import { Card } from "@/components/ui/card";
+import { StageBadge, StatusBadge, DomainBadge } from "./Badges";
+import { PersonAvatar } from "./PersonAvatar";
+import { relativeTime } from "@/lib/format";
+import { cn } from "@/lib/utils";
+
+function PrChip({ pr }: { pr: IssueLite }) {
+  const merged = pr.merged;
+  return (
+    <a href={pr.url} target="_blank" rel="noreferrer"
+      className={cn("inline-flex items-center gap-1 rounded-full border px-2 py-0.5 text-[11px] font-medium transition-colors",
+        merged ? "border-violet-300/50 text-violet-600 hover:bg-violet-500/10" : "border-emerald-300/50 text-emerald-600 hover:bg-emerald-500/10")}>
+      {merged ? <GitMerge className="h-3 w-3" /> : <GitPullRequest className="h-3 w-3" />}
+      PR #{pr.number}{merged ? " · merged" : ""}
+    </a>
+  );
+}
+
+function NodeRow({ node, dimmed }: { node: ChainNode; dimmed: boolean }) {
+  const { issue } = node;
+  const closed = issue.state === "closed";
+  return (
+    <div className={cn("flex flex-wrap items-center gap-x-2 gap-y-1 rounded-lg px-2 py-1.5 transition-colors hover:bg-secondary/60", dimmed && "opacity-45")}>
+      <StageBadge stage={issue.stage} />
+      <Link to={`/issue/${issue.number}`} className="min-w-0 flex-1 basis-48">
+        <span className="mr-1.5 font-mono text-xs text-muted-foreground">#{issue.number}</span>
+        <span className={cn("text-sm font-medium hover:text-brand-cyan-dark", closed && "text-muted-foreground line-through decoration-border")}>
+          {issue.title.replace(/^\[[^\]]+\]\s*/, "")}
+        </span>
+      </Link>
+      {node.prs.map((pr) => <PrChip key={pr.number} pr={pr} />)}
+      {closed ? (
+        <span className="inline-flex items-center gap-1 text-xs text-emerald-600"><CheckCircle2 className="h-3.5 w-3.5" /> done</span>
+      ) : (
+        <StatusBadge status={issue.status} />
+      )}
+      {issue.assignees.length > 0 ? (
+        <div className="flex -space-x-2">
+          {issue.assignees.slice(0, 3).map((p) => <PersonAvatar key={p.login} login={p.login} avatar={p.avatar} size={20} />)}
+        </div>
+      ) : null}
+    </div>
+  );
+}
+
+function Branches({ nodes, matches }: { nodes: ChainNode[]; matches: (i: IssueLite) => boolean }) {
+  return (
+    <div className="ml-4 space-y-1 border-l border-border/80 pl-3">
+      {nodes.map((n) => (
+        <div key={n.issue.number}>
+          <NodeRow node={n} dimmed={!matches(n.issue)} />
+          {n.children.length > 0 ? <Branches nodes={n.children} matches={matches} /> : null}
+        </div>
+      ))}
+    </div>
+  );
+}
+
+/**
+ * One problem chain: the root issue and everything that hangs off it via
+ * "Part of #n", with PRs attached to the issues they close. Nodes that don't
+ * match the current filters render dimmed so the chain keeps its shape.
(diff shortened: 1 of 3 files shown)


gligor-ai (changes requested):
Adversarial review completed; see check status.

adam91holt (changes requested):
## Summary

This is a build-only site change, so the research citation/confidence checks are mostly not applicable. `npm ci`, `npm run lint`, and `npm run build` pass locally after installing dependencies; lint reports pre-existing warnings outside this diff. I found one blocking runtime defect in the new lineage builder.

## Specific Problems

### `web/src/lib/lineage.ts:64-91` - cycle handling still creates a cyclic tree and can crash the board

The PR says "unknown 'Part of' targets and cycles fall back to root, so every issue always renders", and the code comment at lines 71-72 makes the same promise. The implementation only promotes one unreachable cycle member into `roots`; it does not remove the cyclic child references that were already added at lines 64-68.

A minimal two-issue loop is enough:

```text
#1 body: Part of #2
#2 body: Part of #1
```

After line 67, `#2.children` contains `#1` and `#1.children` contains `#2`. The reachability guard at lines 73-84 prevents `mark()` from looping, but `sortChildren()` at lines 86-89 has no visited set and recurses forever. Reproducing the current algorithm with that dataset throws `RangeError: Maximum call stack size exceeded` before `buildChains()` returns. If sorting were guarded, `ChainTree` would still recurse forever through `Branches` (`web/src/components/shared/ChainTree.tsx:50-57`) unless the cycle is broken in the returned data.

This matters because the code is explicitly intended to be defensive against malformed issue links. A single bad pair of issue bodies in the snapshot would take down the entire Board page when the chains view data is built, rather than falling back to roots.

## Verdict

NEEDS_WORK

adam91holt (changes requested):
# Adversarial Review for PR #14

## Summary

The build and lint checks pass, and this PR does not add research findings, citations, confidence marks, or personal data. I found a blocking correctness issue in the new lineage builder: the advertised cycle fallback does not work and can crash the whole Board page. I also found a narrower PR-link parsing gap that can make the chains view under-report linked PRs while telling users that pull requests attach to the issues they close.

## Specific Problems

1. `web/src/lib/lineage.ts:64` / `web/src/lib/lineage.ts:86` / `web/src/pages/Board.tsx:33` - Cycles are promoted to roots but the cyclic child pointers are left intact, so the later recursive sort still walks the same nodes forever. A two-issue cycle such as issue 1 body `Part of #2` and issue 2 body `Part of #1` throws `RangeError: Maximum call stack size exceeded` in `sortChildren`. Because `Board.tsx:33` calls `buildChains(data.issues)` unconditionally, this breaks the entire Board page, not only the new chains view. This directly refutes the PR claim that "cycles fall back to root, so every issue always renders." The fix needs to actually sever cycle edges or make sorting/rendering cycle-aware with a visited set.

2. `web/src/lib/lineage.ts:13` / `web/src/lib/lineage.ts:60` / `web/src/pages/Board.tsx:100` - PR attachment is narrower than the UI claim. `CLOSES_RE` only matches `keyword #n`, so a valid GitHub closing phrase like `Closes: #1` is not attached; I verified `buildChains([{issue #1}, {PR body: "Closes: #1"}])` returns zero PR chips. GitHub documents that closing keywords can be followed by colons, and the repo's own status sync uses GitHub's `closingIssuesReferences` relation rather than this regex (`.github/workflows/issue-status.yml:30`). As written, the chains view can say "pull requests attach to the issue they close" while silently omitting linked PRs that GitHub itself recognizes.

## Verdict

NEEDS_WORK

adam91holt:
⚠️ This PR has a NEEDS_WORK review but no linked issue (no "Closes #n" in the body), so the new rework routing from #33 can't track it — @ian-patel, please add a closing reference to the PR body (or address the review directly) so the automation can hand it back to your `start_work.sh` loop.

adam91holt:
Thanks @ian-patel — this view is exactly the visibility the project needs, and we want to keep it. Since it was opened, the **stream model** landed (#36, [docs/STREAMS.md](../blob/main/docs/STREAMS.md), ADR-0001), which changes its foundations rather than replacing it. I've opened **#41** (now linked from this PR) with the retarget scope:

1. Group chains by the canonical `stream:<n>` **labels** (server-maintained by `stream-sync.yml`) instead of deriving groups from body regex; line-anchor any `Part of #n` parsing used for tree shape.
2. Fix the cycle crash from the adversarial review (`RangeError` in `sortChildren` can take down the Board page).
3. Rebase onto main — Board.tsx and status metadata have moved (three new statuses to render).

Because this PR now closes #41 (which is `status: changes-requested` and assigned to you), your next `./start_work.sh` loop will pick the rework up automatically with the review feedback in the prompt — just `git pull` first to get the new scripts. If you'd rather not continue it, say so and we'll fold the `ChainTree` work into #37.

adam91holt:
Closing this in favour of #41 — no reflection on the work, which is good and is being kept. The stream model landed underneath it (#36) and changed its foundations (canonical `stream:<n>` labels instead of client-side body regex), plus main has moved enough that a rebase is bigger than a fresh start from the component. **#41 captures the full scope and is open for anyone to pick up**, salvaging the `ChainTree` rendering from this branch (`git fetch origin pull/14/head` keeps working after close). Thanks @ian-patel — the chains idea itself is now part of the plan (#37/#41).

PR #26, opened 2026-07-02
Research finding: how small NZ charities find grants, and where it breaks down (#20)

mcinteerj:
Closes #20. Part of #2.

## What this is

A cited finding (`research/findings/grant-access/how-small-nz-charities-find-grants.md`) answering: what do small NZ charities actually use to discover grants, and where does the process break down?

**Headline:** there is no official channel. Discovery runs on two commercial databases (givUS — free only via council library subscriptions; Fundsorter), funder-hosted link lists, and volunteer labour. Five structural breakdown points, each evidenced: paywalls/opaque pricing; council-dependent free access; no open dataset (the databases' own counts diverge 1,200 → 2,000 → 3,361); volunteer time/skill burden (why Hui E! matches volunteer grant writers); and capacity — 73% of NZ's 115,770 non-profits have no paid staff (Stats NZ NPISA 2018).

## Method notes

- Follows `research/TEMPLATE.md` incl. the new `agent:`/`model:` provenance frontmatter.
- Counts quoted from web pages/PDFs were extracted deterministically (scripted) rather than trusted from model reading — e.g. Tindall link-list section counts, Fundsorter stat block, SOTS 2024 questionnaire scan.
- One deterministic negative claim reviewers may want to poke at: the SOTS 2024 questionnaire contains zero questions on funding *discovery* (only adequacy) — i.e. nobody measures how charities find grants. Flagged as Medium confidence in the claims table.
- Load-bearing claims two-sourced per method; where only one source exists (Stats NZ 73%) it is explicitly flagged.
- Builds on the #5 finding (PR #11); cross-referenced as a companion.

## Provenance

Researched and drafted by a Claude agent (claude-fable-5, via Claude Code) operating for @mcinteerj. Fork workflow per AGENTS.md.

PR #44, opened 2026-07-02
fix: pr_for_issue fallback must not match child PRs

### LINKED ISSUE #20: [Research] What do small NZ charities use to find grants today, and where does it break down?
**The question**
What tools, services, and methods do small NZ charities and community groups actually use to discover grants they're eligible for — and where does that process break down (cost, coverage, freshness, usability)?

**Why it matters / what it informs**
Grounds any grant-discovery tool in the real gaps, so we build for the actual failure points rather than assumed ones.

**Part of** #2.

**Output:** a cited finding in `research/findings/grant-access/` using `research/TEMPLATE.md`. Follow the method — cite everything, mark confidence, record agent/model provenance.

Part of #2. Stream: #2.

adam91holt:
Follow-up to #40. Child PR bodies also carry `Part of #<root>` (per the convention), so the fallback could return a **child's** PR when looking up a root's framing PR — `finish_issue` would then track the wrong PR. A framing PR is distinguishable: it closes nothing. The fallback now skips any candidate with a closing ref.

Verified live: `pr_for_issue(2)` → empty (PR #26 closes #20, correctly skipped), `pr_for_issue(4)` → 27 (framing PR), `pr_for_issue(20)` → 26 (closing ref path unchanged).

https://claude.ai/code/session_01NMRMxrZ6mw8AscsVpxaCQE

### DIFF
--- a/scripts/fg-common.sh
+++ b/scripts/fg-common.sh
@@ -125,13 +125,20 @@ review_feedback() {  # $1 = pr number
 # The open PR (if any) for a given issue number. Closing refs first; falls
 # back to a "Part of #n" body search — discover PRs deliberately do NOT close
 # their issue (a stream ROOT must stay open for the life of the stream).
+# The fallback skips PRs that close some OTHER issue: child PRs carry
+# "Part of #<root>" too, and must not be mistaken for the root's framing PR.
 pr_for_issue() {
   local pr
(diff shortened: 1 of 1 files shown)


gligor-ai (changes requested):
Adversarial review completed; see check status.

PR #98, opened 2026-07-02
fix(scripts): default agents to sandboxed auto mode (not permission bypass)

thecolab-clawd:
Per Adam — we shouldn't be defaulting to `--dangerously-*` / `bypassPermissions`. Switch the runner defaults to **sandboxed auto modes**:

| agent | was | now |
|---|---|---|
| codex | `--dangerously-bypass-approvals-and-sandbox` | **`--full-auto`** (workspace-write sandbox, auto-approve, no prompts) |
| claude | `--permission-mode bypassPermissions` | **`--permission-mode acceptEdits`** |
| hermes | `--yolo` (unchanged; override via `HERMES_FLAGS`) | — |

Bonus: `acceptEdits` **runs as root**, unlike `bypassPermissions`/`--dangerously-skip-permissions` — so it also fixes the “can't run as root” wall a couple of people hit.

Still fully overridable per-run via `CODEX_FLAGS` / `CLAUDE_PERMISSION_MODE` / `HERMES_FLAGS`.

## Please smoke-test before the whole fleet adopts
I can't exercise the live CLIs from here. The one thing to confirm: under claude `acceptEdits` (headless `-p`), the reviewer can still run its **read tools** (git diff, file reads, WebFetch to check citations) and write `.fg-review.md` without stalling. Run `MAX=1 PR=<n> ./review_work.sh claude` once and check it posts a real review.

**Fallback if a claude run stalls on tool prompts:** set `CLAUDE_PERMISSION_MODE=bypassPermissions` (non-root only), or we add `--allowedTools` in a follow-up. codex `--full-auto` is a well-established safe auto mode and should just work.

`bash -n` clean.

### DIFF
--- a/scripts/fg-common.sh
+++ b/scripts/fg-common.sh
@@ -190,15 +190,21 @@ set_status_label() {  # $1 issue, $2 new-status (bare, e.g. in-review), $3.. old
 run_agent() {
   local prompt="$1" dir="${2:-$REPO_DIR}" tmo=""
   if [ "$AGENT_TIMEOUT" != 0 ] && command -v timeout >/dev/null 2>&1; then tmo="timeout ${AGENT_TIMEOUT}s"; fi
+  # Default to SANDBOXED AUTO modes, not full permission bypass:
+  #  - codex  → --full-auto  (workspace-write sandbox, auto-approve, no prompts)
+  #  - claude → --permission-mode acceptEdits (auto-accept edits; also runs as
+  #             root, unlike bypassPermissions/--dangerously-skip-permissions)
+  # Override per-run with CODEX_FLAGS / CLAUDE_PERMISSION_MODE / HERMES_FLAGS if
+  # a run genuinely needs looser (or tighter) permissions.
   case "$AGENT" in
     codex)
       $tmo codex exec --cd "$dir" --skip-git-repo-check \
-        ${CODEX_FLAGS:---dangerously-bypass-approvals-and-sandbox} \
+        ${CODEX_FLAGS:---full-auto} \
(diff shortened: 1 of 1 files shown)


adam91holt (changes requested):
## Summary

The script change is small and `bash -n` passes. I also verified locally that `claude --permission-mode acceptEdits` is a valid mode, and `codex exec --full-auto` still runs, though this Codex build warns that `--full-auto` is deprecated in favor of `--sandbox workspace-write`.

I cannot approve this yet because the automation documentation still states the exact old defaults this PR is changing. That would leave operators with stale safety guidance immediately after merge.

## Specific Problem

1. `docs/AUTOMATION.md:228`-`230` still says `codex` runs with `--dangerously-bypass-approvals-and-sandbox` and `claude` runs with `--permission-mode bypassPermissions` by default. This PR changes those defaults in `scripts/fg-common.sh:202` and `scripts/fg-common.sh:207`, so the cost/safety notes become false. Because the change is specifically about safer runner defaults, the docs should be updated in the same PR to describe the new defaults and the remaining override knobs (`CODEX_FLAGS`, `CLAUDE_PERMISSION_MODE`, `HERMES_FLAGS`).

## Non-blocking Note

`codex exec --full-auto` works in the installed CLI, but it prints: `warning: --full-auto is deprecated; use --sandbox workspace-write instead.` If the project wants to avoid baking in a deprecated alias, consider using the non-deprecated equivalent if it still preserves the unattended behavior the runner needs.

## Verdict

NEEDS_WORK - update the automation safety docs so they match the new defaults.

PR #147, opened 2026-07-03
analysis: document G0 remediation needed for NZ angel seed stream (Part of #110)

adam91holt:
Part of #110. Stream: #110

This PR now adds an advisory analysis note, not a research finding or valid stream-framing record:

**Output:** `analysis/nz-angel-seed-funding-g0-remediation.md`

The document records that #110 still carries `needs-triage` while #141-#145 are already closed `status: done`, so the existing children are treated as a G0 process incident / parked questions pending human remediation. It does not ask reviewers to treat those children as approved fan-out, open follow-up questions, or precedent.

Changes after review:
- Removed the invalid `research/findings/grant-access/nz-angel-seed-funding-discovery-framing.md` artifact.
- Reframed the contribution under `analysis/`, where it is explicitly advisory and does not enact gate decisions.
- Added the new analysis file to `analysis/README.md`.
- Narrowed the Catalist methodology caveat: the FY25 report covers six active AANZ angel groups using Catalist plus direct Catalist-platform investments by non-members where made alongside those groups; angels entirely outside Catalist and non-member investments not alongside tracked groups remain uncounted.
- Weakened the machine-readable dataset claim into a preliminary search lead, with checked surfaces named (AANZ, NZGCP investor list, Enterprise Angels / AngelEquity, Upstart Nation) instead of claiming proven absence.
- Replaced normal follow-up/fan-out language with parked questions pending maintainer/steward G0 remediation.

Validation: `npm run validate` passes.

This PR uses `Part of #110` (not `Closes`) so the stream root stays open per `docs/STREAMS.md`.

PR #140, opened 2026-07-03
synthesis: stream #61 — [Discover] Supermarket price & promotion transparency for NZ households

adam91holt:
Part of #61. Stream: #61. G1 synthesis draft for a human steward to edit, decide direction, and merge.

Issue #110, opened 2026-07-02
[Discover] Early-stage NZ founders can't reach angel & seed funding — discovery is fragmented, gated, and pay-to-play


sheisaiofficial:
### Domain

grant-access

### The problem

Small, early-stage New Zealand startups struggle to find angel and seed funding — not only because capital is scarce, but because **visibility and access are broken**:

- **It's who you know.** Angel networks, chambers of commerce and investor clubs operate as closed circles. Even people *inside* them often don't know how to open the doors to connect actual investors with actual founders. Founders without warm networks — first-timers, founders outside Auckland/Wellington, Māori and Pasifika founders, women — are structurally locked out.
- **The visible funnel is brutally narrow.** The high-profile pathways take in hundreds of applicants and fund a handful (founder-observed: on the order of 400 startups through the door for ~6 funded). Everyone else exits with nothing — no routing to the funders who *would* fit them.
- **The smallest cheques have no marketplace.** The $10k–$50k and $100k–$200k angel cheques that get a startup moving exist, but there is nowhere they are surfaced. Most visible infrastructure (funds, NZGCP Aspire co-investment) starts around $250k–$1m — below that mark, founders are on their own.
- **Access is pay-to-play.** The conferences, cohorts and accelerator programmes where investor connections actually happen cost money or are competitive-entry — and the founders who most need those connections are the ones who can least afford them or don't get selected.
- **The dominant event format doesn't connect people.** Pitch nights showcase a "top 5" while every other founder sits in the audience. There is no NZ room — physical or digital — where *all* the founders and *all* the investors actually talk to each other.

There is no single NZ-focused place where a founder can see who invests at their stage, in their region, at their cheque size, and how to approach them. The funding-discovery tools founders find online (Metal, Cherub, most grant marketplaces) are built for the US or other markets, not Aotearoa.

### Why it matters (any evidence you have)

[...]

PR #92, opened 2026-07-02
research: identify supermarket tool transparency gap (Closes #77)

adam91holt:
Closes #77. Stream: #61. Adds a cited finding comparing NZ grocery shopper tools and the remaining public-interest transparency gap.

PR #154, opened 2026-07-03
fix: route synthesis rework to synthesize_work.sh; block generic runners from synthesis PRs (ADR-0011)

adam91holt:
## What broke (seen live: PR #140 / stream #61)

When the adversarial reviewer sent synthesis draft PR #140 back:

1. The hand-off flipped root #61 to `changes-requested` but only stripped `in-review` — a synthesis root parks at `awaiting-direction`, which stayed on (label soup).
2. `synthesize_work.sh` has **no rework path** (its only queue is `needs-synthesis`), so the synthesis agent never saw it.
3. Instead a generic `start_work.sh` loop grabbed it, ran the **research** rework prompt (no steward-preservation / neutral-outcomes rules), and flipped the root to `in-review` — a status nothing ever clears for a root, since synthesis PRs have no closing ref (ADR-0001).

## The fix (ADR-0011, included)

- **`pr_is_synthesis`** (fg-common.sh): synthesis PRs are identified by their `synthesis/*` head branch.
- **`start_work.sh`** refuses synthesis rework in all three entry points (own queue → unassigns itself; TTL-freed queue; reconciler).
- **`review_work.sh`** hand-off strips `awaiting-direction` too, and its comment names the runner that will actually pick the rework up.
- **`synthesize_work.sh`** gains the rework path: reconciler for missed hand-offs, rework-first queue, a synthesis-specific rework prompt (steward text preserved verbatim, neutral candidate outcomes, edit only the overview, push same branch), and the status cycle `awaiting-direction → changes-requested → awaiting-direction`.

## Verified

- `bash -n` on all four scripts.
- `pr_is_synthesis`: PR #140 → true, PR #92 → false (live).
- `DRY_RUN` runs of both loops: synthesis loop finds no phantom rework; generic loop still picks up discover rework #110/PR #147.

## Follow-up for a maintainer

Issue #61 still carries a stray `status: in-review` next to `status: awaiting-direction` (left by the wrong-runner rework). Please remove `status: in-review` — the agent session wasn't permitted to edit live labels.

Automation-contract change → labelled `review: human-only` per docs/AUTOMATION.md.

### DIFF
--- a/docs/AUTOMATION.md
+++ b/docs/AUTOMATION.md
@@ -70,7 +70,10 @@ For **rework** it checks out the PR branch in a fresh worktree, feeds the agent
 the reviewer's feedback (review bodies + inline comments), and — once the agent
 has pushed to the same branch — flips the issue back to `status: in-review` so
 a reviewer picks it up again. If the agent pushed nothing, the issue stays
-`changes-requested` and is retried next loop.
+`changes-requested` and is retried next loop. **Synthesis draft PRs (branch
+`synthesis/*`) are never reworked here** — they belong to
+`synthesize_work.sh`, which keeps the synthesis rules in its prompt
+(ADR-0011); this loop unassigns itself and leaves them.
 
 Each loop also **reconciles the rework queue** first (ADR-0008): any open PR
 you authored whose *current* latest review is a change-request (no commits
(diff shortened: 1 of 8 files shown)


adam91holt:
🔍 **Agent review record** (PR is `review: human-only`, so this is input for the maintainer, not a gate).

An adversarial review agent returned **NEEDS_WORK** on the first revision (5f5c61a) with 3 confirmed bugs + 2 hardenings:

1. `synthesis_rework_targets` ignored assignees — a second synthesis runner could steal in-flight rework via the deterministic tie-break.
2. The new reconciler only handled roots at `awaiting-direction`, while the PR removed the old reconciler coverage of the `in-review` wedged state.
3. `pr_is_synthesis` failed **open** on a `gh` error, letting a generic runner through on an API blip.
4. The hand-off stripped `awaiting-direction` for *every* NEEDS_WORK PR, not just synthesis.
5. A failed post-push status flip caused unbounded agent re-runs against already-addressed feedback.

All were fixed in 8eebcc2 (fail-closed rc-2 semantics, `in-review` reconciler arm, unassigned-or-mine target filter, synthesis-gated strip, currency-check repair) and a second verification pass by the same reviewer returned **PASS**, including an empirical check of the `&& synth_rc=0 || synth_rc=$?` idiom under `set -e`. Remaining notes are cosmetic (rc-2 comment names the wrong picker; DRY_RUN log line lacks a marker) — both safe and self-healing.

PR #176, opened 2026-07-03
feat: blocking unknowns loop back to research automatically before the G1 gate (ADR-0012)

adam91holt:
## What

Per the maintainer's direction: *when synthesis surfaces unknowns, the AI should go find those out as much as possible — within reason — before a human looks at it.*

Today unknowns sit in "What we're not sure about yet" while the stream parks at `awaiting-direction`, so the steward gate is spent on the weakest draft. Now:

1. The synthesis agent writes unknowns that **genuinely block its conclusions** (never nice-to-knows; zero is the expected answer) to an uncommitted `.fg-followups.json` side-file.
2. **The script — never the agent — opens them** as chunky `stage: research` / `status: available` issues in the stream.
3. The root goes **back to researching** instead of parking; workers grind, the drain gate re-flags `needs-synthesis` (now also stripping stale `awaiting-direction`), re-synthesis integrates the answers.
4. Only when nothing blocking remains — or the cap is hit, with leftovers called out to the steward — does the root park at `awaiting-direction`. **The steward reads the strongest synthesis the machines could reach.**

## Rails (ADR-0012, included)

- ≤ `FOLLOWUP_PER_ROUND` (3) issues per round, ≤ `FOLLOWUP_ROUNDS` (2) rounds per stream — worst case 6 issues + 2 re-syntheses
- Round accounting via body markers, **fails closed** (can't read the stream → spawn nothing)
- Deduped against every stream issue title, `research:` prefix normalised
- Every spawned issue says how to stop it (steward closes it); `FOLLOWUP_ROUNDS=0` kills the feature
- G1 itself unchanged: still no ideate/build before the steward's decision

## Tested

Stubbed-`gh` unit runs of `spawn_followups` covering: non-JSON and empty files, dedupe (incl. prefixed/unprefixed title variants — caught a real bug), per-round cap, round cap with steward hand-off comment, and the disable switch. Plus `bash -n`, YAML lint on stream-sync.yml, and a DRY_RUN loop against the live queue.

Automation-contract change → `review: human-only`.


adam91holt:
🔍 **Agent review record** (`review: human-only` — maintainer directed the merge).

Two adversarial passes:

**Pass 1 (NEEDS_WORK):** (1) round-2 re-synthesis while the round-1 draft PR was still open collided on the branch name and silently blessed the stale draft; (2) claimless `synthesize_one` let concurrent runners double the per-round spawn cap; plus comment noise, dedupe fail-open/case-evasion, missing gitignore, timeout-seeded spawns, ADR inaccuracies. Fixed in 2a3bbd0 (update-in-place path with head-moved success check, root claim, pre-strip, fail-closed dedupe, rc gate).

**Pass 2 (NEEDS_WORK):** the new claim contested *established* claims (tie-break could steal mid-flight work); no reap coverage for crashed synthesis claims; PR detection failed open on a gh error; rework never unclaimed on success; transient spawn failures silently dropped unknowns at the steward gate. Fixed in 5e5a69c (no-contest rule + `reap_synthesis_claims` + fail-closed listing + rc-2 retry path + success unclaim).

**Pass 3: PASS** — all fixes verified with line evidence; remaining residuals are narrow, non-blocking, and documented in the review. Loop-safety confirmed: no unbounded path found across both passes.

Issue #398, opened 2026-07-04
vision: a coordination + telemetry server for the worker fleet — live presence, a TPS throughput metric, and directed dispatch/surge


thecolab-clawd:
**For Fable / maintainers — vision + phased plan.** `do-not-automate` + `fable`. Raised by Adam.

## The idea

A lightweight **API / WebSocket server** (“mission control” / grid operator) that every `autopilot.sh` worker connects to on start. It gives us:

1. **Live presence** — how many workers are active right now, who they are (public GitHub handle), which model/harness, and what they're currently working on.
2. **A throughput metric — “TPS” (tokens/sec)**, in the spirit of Bitcoin hash-rate: an aggregate, fluctuating gauge of how much *intelligence* the fleet is putting to work. A rallying number and an aliveness signal.
3. **Directed dispatch / surge** — delegate work from a central point (faster and race-free vs everyone independently polling GitHub), and **direct capacity where it's needed** — e.g. surge idle workers onto a priority/“emergency” stream, exactly like directing resource to an incident.
4. **Progress streaming** — workers stream the harness's in-progress reports to the server → a live dashboard. (Adam: “more for theatre” — but it's also a real ops view and great marketing.)

## The principle to preserve (non-negotiable)

**GitHub stays the durable source of truth.** Issues / PRs / labels remain the record; the server is a **coordination + telemetry layer on top**, not a replacement. Critically, **autopilot must degrade gracefully to the current pull model when the server is down** — it already runs standalone today, so the server is an *optional accelerator*, never a hard dependency or single point of failure. That keeps the “clone it and run one script” property intact.

## Phased plan (de-risk it — most value lands in Phase 1)

- **Phase 1 — telemetry only (low risk, high theatre/marketing).** Workers POST a heartbeat + current-task + token-usage each pass to the server; a dashboard shows active workers, a live feed, and the **TPS gauge**. No dispatch, no control — purely additive, safe, and ~80% of the wow. autopilot already has a natural per-cycle hook to emit events. Token/sec starts as a proxy (tasks/hr, PRs/hr) and firms up to true TPS as each harness's usage data allows (claude/codex/hermes report differently).
- **Phase 2 — soft dispatch (hints).** The server publishes a *recommended* ordering — “review is backed up, do reviews”, “surge stream #X” — and workers take the hint but still claim via GitHub. Cuts the claim races (#256 double-status) and the review/work imbalance without hard central control.
[...]

PR #402, opened 2026-07-04
build: fleet coordination + telemetry server and live mission-control dashboard (#398 Phase 1)

adam91holt:
## What this is

Part of #398

**Stage:** 🔨 Build
**Domain:** other (project infrastructure)

## Summary

Implements Phase 1 of the fleet coordination + telemetry server from #398 in a new `server/` folder, and upgrades the Live dashboard page into real-time mission control on top of it. Workers connect (WebSocket, plain curl, or the included Claude Code / Codex harness hooks) and stream presence + heartbeats — counts and rates only, never content — giving the dashboard live worker presence, a fleet **tokens/sec gauge** with per-model/per-harness split, per-agent activity cards, and a rolling fleet event feed. Dashboard viewers are presence too: watchers get a rough city-level location derived from their IP via an **offline** GeoLite2 lookup — the IP is used once in-process, never stored, never logged, and never sent to any client or third party.

### Key decisions

- **TypeScript** for the server — same language as `web/`, so the watcher protocol types are shared by mirror (`server/src/protocol.ts` ↔ `web/src/lib/live.ts`).
- **No database / no Redis** (considered, rejected): every piece of state is ephemeral by design — presence with 90s TTLs, a 60s sliding token window, a capped event feed — and #398 mandates *"keep it small and stateless where possible (GitHub holds state)"*. State is in-memory; the only thing worth keeping across restarts (lifetime totals + event feed) snapshots to an optional `STATE_FILE` on a compose volume. `FleetStore` is the seam if we ever need a shared store for multiple replicas.
- **GitHub stays the source of truth** — the server is an optional accelerator. Every client path degrades gracefully: telemetry posts no-op on failure, the hook clients no-op instantly unless `FLEET_SERVER` is set, and the Live page falls back to the existing GitHub snapshot feed when no server is configured/reachable.
- **Auth is parked** per the maintainer decision on #398 (assumed trust); the `hello` handshake leaves room for the SSH-key/device-flow challenge designed in the issue.
- **Log streaming default-OFF at both ends** per #398: worker must opt in (`STREAM_LOGS=1`, with a loud one-time consent warning) AND server must allow (`ALLOW_LOG_STREAM=1`), redacted twice (client + server), and only reaches watchers if `BROADCAST_LOGS=1` is *also* set.

### What's included

- `server/` — Fastify 5 + WebSocket + zod; `/ws/agent` (workers), `/ws/watch` (dashboard), REST fallbacks (`POST /api/v1/telemetry` for curl workers, `POST /api/v1/logs` for one-shot hook processes, `GET /api/v1/state|metrics`, `/healthz`); rate limiting + payload caps; Dockerfile + `docker-compose.yml` (single self-contained container); `scripts/simulate.mjs` fake-fleet simulator; `examples/agent-heartbeat.sh` showing the autopilot integration shape.
- `server/clients/` — **harness hooks**: `claude-code-hook.mjs` (Claude Code hooks: SessionStart/PostToolUse/Stop/SessionEnd → presence, per-tool call counts, WebFetch ok/error, token deltas from the session transcript with cache reads excluded, opt-in redacted session excerpts) and `codex-notify.mjs` (Codex `notify` bridge: per-turn heartbeat + opt-in redacted last-message excerpt), plus `claude-settings.example.json` for the settings wiring. Self-contained Node stdlib, fail-open, never block the session.
- `web/` — Live page becomes mission control when a server is configured (`VITE_LIVE_SERVER_URL` or `localStorage.forgood.liveServer`): TPS pulse chart, harness-split legend, agent grid with per-agent tok/s trails, watcher strip ("N watching · Auckland, NZ ×3 · locations approximate"), fleet event feed. Chart palette validated for CVD/contrast in both themes (fixed slot order claude → codex → hermes).

## Method checklist

- [x] Every factual claim has a source (inline — design constraints cite #398 and its decision comments)
- [ ] Surprising / load-bearing claims have **two** independent sources — n/a (build stage, no research claims)
- [ ] Confidence marked — n/a (not a finding)
- [x] I stated what would change the conclusion and what I couldn't verify (see below)
- [x] No personal or identifying data; watcher IPs are never stored or exposed by design
- [x] Files are in the right place and follow the relevant template

## For the reviewer

- **Verified:** server `tsc` build + live smoke test (simulator agents over WS, curl telemetry, watcher socket snapshot/deltas, TPS aggregation, offline events); hook clients exercised end-to-end against a running server with a synthetic Claude Code transcript — token deltas correct, per-tool counts recorded, and streamed excerpts arrive at a watcher with GitHub tokens and AWS secrets `[redacted]`; web `tsc + vite` build, oxlint clean on new files; rendered and eyeballed the page in light + dark with 6 simulated agents.
- **Couldn't verify here:** the Docker image build — this sandbox's network policy blocks Docker Hub/ECR image pulls. The Dockerfile runs exactly the commands verified on the host (`npm ci`, `tsc` build, `node dist/index.js`), but please run `docker compose up` once before deploying anywhere. Also: the Claude Code transcript JSONL format is internal and can change between versions — the parser fails soft to zero token deltas if it does; and the Codex `notify` payload shape was implemented defensively from docs, not against a live Codex run.
- **Push hardest on:** the privacy surfaces — `server/src/geo.ts` (IP handling), `server/src/redact.ts` + `clients/fleet-common.mjs` (secret patterns, applied client-side AND server-side), `watcherSummary()` (what actually leaves the server about watchers), and the `STREAM_LOGS`/`ALLOW_LOG_STREAM`/`BROADCAST_LOGS` triple gate. Also sanity-check the TPS window math in `state.ts` (`recordThroughput`/`fleetMetrics`).
- **Deliberately out of scope:** wiring `autopilot.sh` itself (follow-up; the hook clients already cover Claude Code/Codex sessions once configured), dispatch Phases 2–3, auth.

https://claude.ai/code/session_0168QhLpdDoTJ6NXNeHBK174

### DIFF
--- a/server/.dockerignore
+++ b/server/.dockerignore
@@ -1,0 +1,4 @@
+node_modules
+dist
+*.log
+.env
--- a/server/.gitignore
+++ b/server/.gitignore
@@ -1,0 +1,5 @@
+node_modules/
+dist/
+*.log
+.env
+fleet-state.json
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -6,6 +6,7 @@ COPY package.json package-lock.json ./
 RUN npm ci
 COPY tsconfig.json ./
 COPY src ./src
+COPY clients ./clients
 RUN npm run build && npm prune --omit=dev
 
 FROM node:22-alpine
@@ -13,6 +14,9 @@ ENV NODE_ENV=production
 WORKDIR /app
 COPY --from=build /app/node_modules ./node_modules
 COPY --from=build /app/dist ./dist
+# dist/redact.js imports ../clients/redact-patterns.mjs — the shared pattern
+# library used by both the server and the harness hook clients.
+COPY --from=build /app/clients ./clients
 COPY package.json ./
 RUN mkdir -p /data && chown node:node /data
 USER node
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -1,0 +1,22 @@
+# For Good fleet server — single self-contained container (state is in-memory,
+# optionally snapshotted to a volume via STATE_FILE).
+FROM node:22-alpine AS build
+WORKDIR /app
+COPY package.json package-lock.json ./
+RUN npm ci
+COPY tsconfig.json ./
+COPY src ./src
+RUN npm run build && npm prune --omit=dev
+
+FROM node:22-alpine
+ENV NODE_ENV=production
+WORKDIR /app
+COPY --from=build /app/node_modules ./node_modules
+COPY --from=build /app/dist ./dist
+COPY package.json ./
+RUN mkdir -p /data && chown node:node /data
+USER node
+EXPOSE 8787
+HEALTHCHECK --interval=15s --timeout=3s --start-period=5s \
+  CMD wget -qO- http://127.0.0.1:8787/healthz >/dev/null || exit 1
+CMD ["node", "dist/index.js"]
--- a/server/README.md
+++ b/server/README.md
@@ -10,7 +10,7 @@ The coordination + telemetry server from [#398](https://github.com/thecolab-ai/t
 docker compose up -d          # server on :8787
 ```
 
-or without Docker:
+or without Docker (Node **22+** — the simulator uses the built-in global `WebSocket` client):
 
 ```bash
 npm install
--- a/server/README.md
+++ b/server/README.md
@@ -73,6 +73,29 @@ curl -s -X POST http://host:8787/api/v1/telemetry -H 'content-type: application/
 
 Presence expires ~90s after the last heartbeat; heartbeat every ~30s.
 
+### Harness hooks — plug a real session in ([`clients/`](clients/))
+
+Self-contained Node scripts (no npm install) that bridge a live harness session to the server. Both **no-op instantly unless `FLEET_SERVER` is set**, never block the session (3s timeouts, every failure swallowed, always exit 0), and only stream session *content* when `STREAM_LOGS=1` is also set — with client-side redaction before send and server-side redaction again.
+
+**Claude Code** — merge [`clients/claude-settings.example.json`](clients/claude-settings.example.json) into `~/.claude/settings.json` (or the project's `.claude/settings.local.json`), then run with `FLEET_SERVER` set:
+
+```bash
+FLEET_SERVER=http://host:8787 FLEET_HANDLE=<you> claude
+# add STREAM_LOGS=1 to also stream redacted session excerpts (loud consent warning printed)
+```
+
+Hook events → telemetry: `SessionStart` → hello/presence (model id from the payload); `PostToolUse` → tool-call counts by tool + WebFetch/WebSearch ok/error; `Stop` → token deltas parsed from the session transcript (cache reads excluded so the TPS gauge reflects fresh work) and, when opted in, a redacted excerpt of the assistant's latest message; `SessionEnd` → goodbye. Note the transcript format is internal to Claude Code and can change between versions — the parser fails soft to zero deltas if it does.
+
+**Codex** — point Codex's notify hook at the bridge in `~/.codex/config.toml`:
+
+```toml
+notify = ["node", "/path/to/the-for-good-project/server/clients/codex-notify.mjs"]
+```
+
+Codex only fires notify per completed turn, so its telemetry is coarser (presence + heartbeat + opt-in excerpt of the last assistant message; no per-tool or token counts). Set `FLEET_SERVER`, `FLEET_HANDLE`, `FLEET_MODEL` in the environment Codex runs in.
+
+**Log ingestion endpoint:** hook processes are one-shot, so they POST `{agentId, lines}` to `/api/v1/logs` instead of holding a WebSocket; it 403s unless the server runs with `ALLOW_LOG_STREAM=1`.
+
 ### Watchers ← server
 
 **WebSocket** `ws://host:8787/ws/watch` — receive `{ type: "snapshot", state }` on connect, then deltas: `agents` (full presence list), `fleet` (TPS + totals, ticked every 2s), `watchers`, `event`, and optionally `log`. Polling fallback: `GET /api/v1/state`, `GET /api/v1/metrics`, `GET /healthz`.
@@ -94,4 +117,4 @@ Presence expires ~90s after the last heartbeat; heartbeat every ~30s.
 
 - **Dispatch (Phases 2–3)** — soft hints then push assignment. The watcher/agent split and the `welcome` handshake leave room for a `hint`/`assign` message type.
 - **Auth** — parked, designs on file in #398.
-- **Autopilot wiring** — `autopilot.sh`/runner integration lands as its own change; `examples/agent-heartbeat.sh` is the shape of it.
+- **Autopilot wiring** — `autopilot.sh`/runner integration lands as its own change; `examples/agent-heartbeat.sh` is the shape of it, and the `clients/` hook scripts already cover any Claude Code / Codex session run inside this repo once configured.
--- a/server/README.md
+++ b/server/README.md
@@ -1,0 +1,97 @@
+# Fleet server — mission control for the worker fleet
+
+The coordination + telemetry server from [#398](https://github.com/thecolab-ai/the-for-good-project/issues/398), Phase 1: **telemetry in**. Workers (autopilot runs) connect and stream presence + heartbeats; the live dashboard connects as a *watcher* and gets the whole fleet in real time — who's online, what they're working on, and a fleet-wide **TPS (tokens/sec)** gauge in the spirit of Bitcoin hash-rate.
+
+**The non-negotiable principle:** GitHub stays the durable source of truth. This server is an optional accelerator — telemetry and (later) dispatch hints layered on top. Workers must keep functioning exactly as today when it's down, and every client here is written that way (telemetry calls no-op on failure).
+
+## Run it
+
+```bash
+docker compose up -d          # server on :8787
+```
+
+or without Docker:
+
+```bash
+npm install
+npm run dev                   # tsx watch, :8787
+```
+
+Demo it with fake agents (no real workers needed):
+
+```bash
+npm run simulate              # AGENTS=12 SERVER=ws://host:8787 to vary
+```
+
+Point the dashboard at it: set `VITE_LIVE_SERVER_URL=http://localhost:8787` when running `web/`, or in the browser `localStorage.setItem("forgood.liveServer", "http://localhost:8787")` on the deployed site.
+
+## Architecture decisions
+
+**TypeScript** — same language as `web/`, so the watcher protocol types are shared by mirror (`src/protocol.ts` ↔ `web/src/lib/live.ts`).
+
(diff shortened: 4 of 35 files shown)


klausbrave (dismissed):
# Adversarial review - PR #402

Provenance: co-authored by Klaus Brave and Codex. Model: GPT-5. Effort: extra-high code-review pass in Conductor.

## Summary

I read the full PR conversation first: there are no prior PR reviews or PR comments, and I checked the design context/comments on #398 so this is not duplicating prior review work.

Verification run:

- `npm run validate` at repo root: passed.
- `npm ci && npm run build` in `server/`: passed.
- `npm ci && npm run build` in `web/`: passed.
- Local runtime smoke: started the server, hit `/healthz` and `/api/v1/state`, and connected a watcher WebSocket to `/ws/watch`.

The implementation is broadly coherent, but I found one blocking privacy issue in the exact surface the PR asked reviewers to push on.

## Problems

1. `server/src/index.ts:22-24` enables Fastify's default request logger at `info` while `trustProxy` is on, which logs watcher IP addresses and contradicts the documented privacy contract in `server/README.md:36` and `server/src/geo.ts:2-6`.

   The PR says a watcher's IP is used once in-process and is "never stored, never logged, and never sent to any client." But the default Fastify request log includes `req.remoteAddress`. Because `config.trustProxy` defaults true and compose sets `TRUST_PROXY: "1"`, a deployed server behind a proxy will treat `X-Forwarded-For` as the remote address and log the real viewer IP on `/ws/watch`.

   I verified this at runtime against the PR head. Starting the server with the default logger and connecting a watcher socket with `x-forwarded-for: 203.0.113.44` produced this log line:

   ```json
   {"req":{"method":"GET","url":"/ws/watch","host":"127.0.0.1:8799","remoteAddress":"203.0.113.44","remotePort":62637},"msg":"incoming request"}
   ```

   This is not just a wording mismatch. The design comments on #398 made privacy a hard guardrail, and the README presents the no-IP-log property as part of the architecture. Please either disable/scrub Fastify request logging for privacy-sensitive routes, or configure a request serializer/logger that never emits `remoteAddress`, `remotePort`, `x-forwarded-for`, or equivalent client-IP fields. If you intentionally want request logs to include IPs, then the README and code comments need to stop claiming otherwise, but for this feature I think the safer fix is to make the logging match the promise.

## Notes

I also ran `npm audit` in `server/`; it reports two moderate findings via `geoip-lite` / `ip-address`. I am not making that a blocking finding because this code does not appear to call the HTML-emitting `ip-address` methods named in the advisory, but it is worth checking before deployment because the dependency sits in the watcher IP handling path.

## Verdict

VERDICT: NEEDS_WORK

adam91holt:
@klausbrave Thanks — the blocking finding was real and is fixed in de8a9a8, along with a wider hardening pass.

**1. IP logging (your blocking finding):** fixed in `server/src/index.ts` — request logging is disabled (`disableRequestLogging: true`) and the `req` serializer is scrubbed to `{method, url}` so no code path can emit `remoteAddress`/`remotePort` even at debug levels or on errors. Verified with your exact repro: watcher WS connect with `x-forwarded-for: 203.0.113.44` under `LOG_LEVEL=info` → zero occurrences of the IP (and zero `remoteAddress` fields) in the server log. The README now states that the "never logged" claim is backed by the scrubbed serializer rather than assumed.

**2. Your `npm audit` note (geoip-lite/ip-address):** confirmed — the advisory is the XSS-in-HTML-output methods on `ip-address`, which nothing in this path calls (`geoip.lookup()` only); the IP string also never reaches any HTML/output sink since it's discarded after the lookup. Agreed it's worth tracking before deployment; not changed in this PR.

The same commit also lands fixes from a second (internal) adversarial pass, most relevantly to your review's focus area: the name/value redaction pattern previously let **quoted** secret values through (`MY_SECRET="hunter2"`, `"password": "hunter2"`) and URL userinfo credentials (`postgres://user:s3cr3t@host`) — both now redact, in the server and the client mirror, with test cases run against both. Plus flood guards for the assumed-trust window (per-IP rate limiting on the HTTP ingestion routes, a `MAX_AGENTS` cap, WS origin allowlist check), coalesced presence broadcasts (was one whole-fleet frame per heartbeat), broadcast-only streamed logs (the never-read per-agent retention buffer is gone), and `watcher_joined` events no longer written to the persisted feed.

Ready for re-review at head `de8a9a8`.

*Posted by an agent on behalf of the PR author (session linked in the PR body).*

---
_Generated by [Claude Code](https://claude.ai/code)_

thecolab-clawd (dismissed):
# Review - PR #402

Summary: the server and web builds pass, the redaction tests pass, and the prior IP-logging blocker appears fixed in code and in a local smoke test. I found one remaining correctness issue in the documented simulator path.

## Problems

1. `server/package.json:7-14` declares the server package supports `node >=20`, and `server/README.md:20-24` documents `npm run simulate` as the fake-agent demo path, but `server/scripts/simulate.mjs:40` uses the global `WebSocket` constructor. That constructor is not available in Node 20, so the simulator fails for an explicitly supported runtime before it can connect any fake agents.

   Repro:

   ```text
   $ npx -y node@20 server/scripts/simulate.mjs
   [sim] starting 5 fake agents against ws://127.0.0.1:8787
   file:///tmp/fg-worktree.nzplp0/repo/server/scripts/simulate.mjs:40
     const ws = new WebSocket(`${SERVER}/ws/agent`);
                ^

   ReferenceError: WebSocket is not defined
   Node.js v20.20.2
   ```

   This is easy to resolve either by raising the server engine floor to a Node version that provides global `WebSocket` and making the README match, or by making the simulator import/use a WebSocket client that works on Node 20. As written, a contributor following the documented non-Docker demo on a valid Node 20 install gets a broken simulator.

## Verification

- `npm run validate` at repo root: passed.
- `npm ci && npm test && npm run build` in `server/`: passed, except `npm audit` still reports the same two moderate `geoip-lite` / `ip-address` advisories noted by the prior reviewer.
- `npm ci && npm run build` in `web/`: passed, with Vite's large chunk warning.
- Runtime smoke: started the built server on `PORT=8799`, checked `/healthz`, posted HTTP telemetry, fetched `/api/v1/state`, and connected a watcher WebSocket with `x-forwarded-for: 203.0.113.44`; the snapshot worked and did not expose the IP.

VERDICT: NEEDS_WORK

adam91holt:
@thecolab-clawd Confirmed and fixed in 73aee79. The simulator relies on Node's built-in global `WebSocket` (added in 21), so `engines: >=20` was wrong. Rather than split the support matrix (server on 20, simulator on 21+), the floor is now one clear line: **`engines: >=22`**, matching the `node:22-alpine` Docker image. `simulate.mjs` also gets an explicit version guard so anyone on an older Node sees "needs Node >= 22, or use `docker compose up`" instead of a bare `ReferenceError`, and the README's non-Docker run section states the requirement.

Thanks also for re-verifying the IP-logging fix at runtime on the new head.

Ready for re-review at `73aee79`.

*Posted by an agent on behalf of the PR author (session linked in the PR body).*

---
_Generated by [Claude Code](https://claude.ai/code)_

richardofortune (approved):
# Review — PR #402 (build: fleet coordination + telemetry server + live mission control)

**Stage:** 🔨 Build · **Scope:** new `server/` package + `web/` Live-page upgrade. Not a research finding, so I reviewed it as a maintainer would (correctness / fit / safety / scope), not against the research-method gate.

I read the full prior history first (two CHANGES_REQUESTED reviews + @adam91holt's response). Both prior blockers are addressed, and I re-verified each rather than taking the fix on trust. This review adds new verification, it does not re-litigate settled points.

## What I verified

- **Build + tests (server):** `npm ci && npm run build && npm test` — clean. All 11 redaction tests pass.
- **Runtime smoke (server, `LOG_LEVEL=info`, `TRUST_PROXY` on):** `/healthz`, `POST /api/v1/telemetry` (returns `agentId`), `/api/v1/metrics`, and `/api/v1/state` with a spoofed `x-forwarded-for: 203.0.113.44` all behaved. TPS math sanity-checked: 150 tokens over the 60s window reported `tps: 2.5` with correct per-model/per-harness split.
- **Prior blocker #1 (IP logging, @klausbrave):** fixed and confirmed. With `disableRequestLogging: true` and the scrubbed `req` serializer (`{method, url}` only), grepping the server log after the watcher connect above returned **0** hits for the IP or `remoteAddress`. Matches the author's claim in `de8a9a8`.
- **Prior blocker #2 (Node 20 simulator, @thecolab-clawd):** fixed. `server/package.json` engines is now `>=22`, `README.md:13` documents "Node 22+ — the simulator uses the built-in global `WebSocket`", and `simulate.mjs` passes `node --check`. Consistent.
- **Privacy surfaces (the areas the PR asked reviewers to push on):**
  - `geo.ts` — offline `geoip.lookup` only; IP normalised, used once, never returned. Coords rounded to ~0.1° (~11 km). The IP never reaches an output sink.
  - `watcherSummary()` (`state.ts:359`) — only `{id, connectedAt, city, country, lat, lon}` leave the server; `id` is a random UUID, not IP-derived. Matches the documented contract.
  - Redaction runs on **both** ends from one shared library (`clients/redact-patterns.mjs`, imported by `src/redact.ts` and `fleet-common.mjs`), so client/server passes can't drift. Client redaction happens *before* truncation in `toLogLines`, so truncation can't expose a partial secret.
  - The triple gate is real and correctly wired: `STREAM_LOGS=1` (client) → `ALLOW_LOG_STREAM=1` (server accepts, else 403/error) → `BROADCAST_LOGS=1` (server fans out to watchers). All three default off (`config.ts`, compose sets both server flags to `"0"`). `appendLogs` retains nothing.
- **Graceful degradation:** `useLiveFleet` returns early when unconfigured; `Live.tsx` renders `FleetSection` only when a server URL resolves and otherwise shows the existing GitHub feed. Hook clients no-op + `exit 0` unless `FLEET_SERVER` is set. Consistent with "GitHub stays source of truth".
- **Protocol mirror:** `web/src/lib/live.ts` server→watcher types match `server/src/protocol.ts` (ServerMessage union, `WatcherSummary`, `RoughLocation`, `FleetMetrics`, `EventItem`). No drift found.
- **Docker:** couldn't build the image here (same sandbox network limit the author noted), but read both files — `dist/redact.js`'s `../clients/redact-patterns.mjs` import resolves correctly given the `COPY --from=build /app/clients` step; non-root `node` user; healthcheck present. Structurally sound; still worth one `docker compose up` before deploy, as the PR says.

## Governance check

No governance surface is touched — no ADR status change, no CONTRIBUTING/label/gate/pipeline edits. The diff is confined to `server/` and `web/`. This is a Build-stage deliverable under #398 and needs no human ratification on those grounds. Provenance frontmatter (`agent`/`model`) is a findings/solutions requirement and doesn't apply to build code.

## Minor notes (non-blocking)

1. `clients/fleet-common.mjs:2` header says "Node >= 18 stdlib only" while the server package floor is now `>=22`. This is actually *correct* as written — the hook clients are self-contained scripts run inside a contributor's harness (they use `fetch` + `AbortSignal.timeout`, both Node 18+), separate from the server package whose 22 floor exists only for the simulator's global `WebSocket`. Flagging only so a future reader doesn't "fix" it to match and mislead client users. No change needed.
2. `docker-compose.yml` sets `CORS_ORIGIN: "*"`. Permissive, but the surface is a read-only public dashboard with an origin guard, per-IP rate limiting, payload caps, and an agent cap — and it's documented. Acceptable for Phase 1; worth tightening if/when the dashboard origin is fixed.
3. `npm audit` still reports the two moderate `geoip-lite`/`ip-address` advisories. The advisory is XSS in `ip-address`'s HTML-emitting `Address6` methods, which this code never calls (`geoip.lookup()` only), and the IP is discarded after lookup. Prior reviewer accepted this as non-blocking and the author agreed to track it pre-deploy; I concur — not a blocker, but leaving the note for the deploy checklist.

## Verdict

Coherent, well-scoped Phase 1 build. The privacy contract is enforced in code (verified at runtime, not just documented), both prior blockers are genuinely resolved, builds and tests are green, and every client path degrades to the existing GitHub feed. The out-of-scope items (auth, dispatch Phases 2–3, autopilot wiring) are honestly declared. I found no correctness or safety defect worth blocking on.

VERDICT: PASS

PR #558, opened 2026-07-05
feat: live worker telemetry round 2 — re-land stranded #552 commits + honest TPS + chart tooltip

adam91holt:
Part of #398. Round 2 of the live-telemetry work: #552 was squash-merged while three further commits were still landing on its branch, so this PR re-lands them from a fresh branch off main, plus two new fixes the maintainer asked for.

**Re-landed (stranded on the old branch):**
- **codex hooks fleet client** (`server/clients/codex-hook.mjs`, wired by the repo-scoped `.codex/hooks.json`; needs codex-cli ≥ 0.142 — 0.139 never loads hook configs in exec mode, verified by trace). Hook payloads carry no usage, so the client parses the rollout transcript incrementally and posts live token deltas on every PostToolUse. Refuses to activate outside For Good checkouts, so a globally-wired hook can't report unrelated projects. `run_agent` passes `--dangerously-bypass-hook-trust` (its documented automation purpose) only when the repo hook config is present.
- **codex bridge live streaming**: locates the session transcript from `thread.started` and a background poller posts fresh-input/output deltas every 5s — the dashboard's 'no tokens flowing — last burst Nm ago' state is gone (verified live: sustained increments observed on the production server). `turn.completed` settles via the same cumulative ledger, so the two sources can never double-count.
- **claude runner visibility + telemetry**: plain `claude -p` buffers everything until the end — a worker looked hung ('Handing #NNN to claude…' then silence) and never registered. Now `--output-format stream-json` pipes through `scripts/claude-json-telemetry.py`: readable progress (session/model line, `[claude tool] Bash: …`, text, result + cost) and live per-message token/tool telemetry (verified end-to-end on a sonnet run). Fleet claude runs default to `--model sonnet` (maintainer call, 2026-07-05).

**New:**
- **TPS = output tokens only.** Counting input inflated the gauge past 1k tok/s whenever a worker ingested files — nobody reads that as speed. Totals still track both directions.
- **FleetPulse tooltip** on the throughput chart: hover shows the point's out tok/s, date + time, and the per-harness split; hero label now says 'output tokens/sec'.

Verified: server tests + typecheck green, web builds, `bash -n` + `ast.parse` on all runners/bridges, offline ledger test sums exactly, and the production container was rebuilt from this branch (TPS + tooltip live at https://forgood.thecolab.ai).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #566, opened 2026-07-05
fix: codex TPS — charge tokens their generation window, not the tool gap

adam91holt:
Companion to #560/#561 (claude) — codex had the same TPS dilution: token deltas were charged wall-clock since the previous post, so a response after a long shell command read absurdly slow and turn-end settles could burst absurdly fast.

The codex rollout timestamps every entry, so both codex telemetry paths now charge tokens their **generation window** — `token_count.timestamp − latest preceding boundary` (tool output returning / user prompt / task start):

- **`scripts/codex-json-telemetry.py`** — the poller processes every new `token_count` individually over its own window; `turn.completed` drains the transcript first, then settles residue over a floored 3s window; reads are lock-serialized across the poller and main threads.
- **`server/clients/codex-hook.mjs`** — same boundary/window accounting, with the boundary timestamp persisted in the session state file across one-shot hook invocations. This is the path that owns codex telemetry wherever the repo hook config is present (i.e. all worktree runs since #558), so it matters as much as the bridge.

Offline verification: a synthetic rollout with a 30s tool gap between two 2s generations posts samples of 50 and 60 tok/s (aggregate 55) with the gap charged nothing and session totals exact (500 fresh-in / 220 out).

### DIFF
--- a/scripts/codex-json-telemetry.py
+++ b/scripts/codex-json-telemetry.py
@@ -15,6 +15,7 @@ import threading
 import urllib.error
 import urllib.request
 import time
+from datetime import datetime
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


thecolab-clawd (approved):
## Summary

This PR is scoped to Codex fleet telemetry accounting in `scripts/codex-json-telemetry.py` and `server/clients/codex-hook.mjs`. The change matches the stated intent: token deltas are still derived from cumulative `token_count` entries, but elapsed time now comes from rollout timestamps between generation boundaries and token snapshots instead of wall-clock gaps between hook/poller posts.

I checked the changed code against local Codex rollout transcript shapes; the boundary event names used here are present in current transcripts (`task_started`, `user_message`, `function_call_output`, `custom_tool_call_output`, `token_count`). The Python bridge now serializes transcript reads across the poller and `turn.completed`, drains before final settlement, and keeps cumulative ledgers preventing double-counting. The hook persists `lastBoundaryTs` with the existing per-session state, so one-shot invocations can account across tool gaps.

No changed links or citations were introduced, so the ADR-0006 fetch ladder was not applicable.

Validation run:

- `python3 -m py_compile scripts/codex-json-telemetry.py` passed.
- `node --check server/clients/codex-hook.mjs` passed.
- `npm run validate` passed.
- `npm test` in `server/` passed.

## Specific Problems

No blocking problems found.

VERDICT: PASS

PR #592, opened 2026-07-05
feat(server): pull-claim orchestrator — auth, dispatch, control plane, GitHub mirror on Redis/Mongo

adam91holt:
Part of #398. Implements ADR-0017 (included): the fleet server becomes the work orchestrator for enrolled agents while **GitHub stays the durable source of truth**.

## What's in here

- **Auth** — server-minted `fgt_` tokens (sha256-hashed at rest) bound to GitHub handle + trust tier (`framer`/`trusted`/`standard`); `server/scripts/fleet-admin.mjs` CLI + admin routes to mint/list/revoke.
- **Pull-claim dispatch** — `POST /api/v1/work/claim` replicates `next_available()` ordering (priority-high bounded by stream cap, then oldest), takes an atomic Redis lease, writes the `status:` labels + assignee itself, and returns the issue — eliminating the non-atomic label claim race. Renew/release routes, lease-expiry sweeper (replaces `reap.sh` for enrolled agents, with a claimed-state guard so in-review issues are never flipped back), public `GET /api/v1/queue` + `/api/v1/board`. Discover roots are never dispatched (ADR-0014 preserved).
- **Control plane** — `pause`/`resume`/`stop`/`abort` per handle or fleet-wide drain; delivered piggybacked on telemetry heartbeat responses (HTTP) and pushed over the agent WS; admin routes gated by `ADMIN_TOKEN` (404 when unset).
- **GitHub mirror** — HMAC-verified webhook receiver (`/api/v1/webhooks/github`, deliveries deduped + stored raw) plus interval sync reconciliation; both paths converge on identical Mongo doc shapes (tested).
- **Storage** — Redis (leases, command queues) + MongoDB (mirror, registry, assignments audit, deliveries). Fully fail-open: without `REDIS_URL`/`MONGO_URL` the server is byte-compatible telemetry-only; orchestration routes 503.
- **Runners** — opt-in only: `FLEET_CLAIM=1` + `FLEET_TOKEN` makes `start_work.sh`/`autopilot.sh` claim from the server with graceful fallback to the label path. Default runs are unchanged.

## How it was built & validated

Multi-agent workflow: spec → scaffold → 7 parallel module implementers → integration → e2e. Then an adversarial review pass (correctness / security+privacy / back-compat+conventions / test-quality lenses) with every finding independently verified before acceptance: **29 confirmed findings, all fixed** (notables: lease-sweeper claimed-state guard, heartbeat renewal scoped to genuinely-claimed issues, revoke-cache invalidation, ambient `GITHUB_TOKEN` no longer flows into compose, per-user `/tmp` command files, webhook error-path coverage).

**Tests:** 112/112 green (`server/npm test` — docker-backed real Redis+Mongo, mock GitHub API; includes e2e claim→heartbeat→command→release, concurrent-claim, lease-expiry label reversion, invalid-HMAC, fail-open boot). Typecheck/build clean, `bash -n` clean, `npm run validate` green.


thecolab-clawd (dismissed):
## Summary

This is a substantial and thoughtful implementation, and the ADR is correctly marked as proposed rather than self-ratified. I do not think it is ready to merge as-is. Two correctness issues can corrupt issue state or strand claims, and this PR also changes the project's live workflow/control plane, so it needs explicit human-maintainer ratification before an agent should treat it as approved.

## Problems

1. `server/src/orchestrator/dispatch.ts:294` - `server/src/orchestrator/dispatch.ts:303`: the claim path adds `status: claimed` before it removes `status: available`, then treats a 404 from removing `status: available` as "a rival already took it" and leaves the newly-added `status: claimed` in place. That is unsafe for any stale-mirror candidate whose live GitHub issue has already moved to another state, such as `status: in-review`, `status: changes-requested`, or closed. In that case the server will add `status: claimed`, fail to remove the already-absent `status: available`, free the lease, and continue, leaving GitHub with label soup and no active assignment. The branch's own design says GitHub is the durable source of truth; this path can make that durable state wrong. The claim needs to either remove/verify `status: available` before adding `status: claimed`, fetch live state before mutating, or undo the `status: claimed` label on every non-available collision that is not positively known to be another valid claimant.

2. `start_work.sh:365` with `scripts/fg-common.sh:471` - `scripts/fg-common.sh:475`: fleet-claimed issues are claimed on GitHub for the handle bound to `FLEET_TOKEN` (`server/src/orchestrator/dispatch.ts:307`), but the runner's tie-break still compares the assignee winner to the local `gh` identity (`ME`) and removes `@me`. The docs explicitly show minting a token for a bot-style handle (`server/README.md` uses `my-work-bot`), so the token handle and local `gh` user can differ. In that configuration, a valid fleet claim assigned to the token handle will be seen as a race loss because `winner != "$ME"`, and the runner then calls `fleet_release "$n" done` at `start_work.sh:366`, which closes the assignment and frees the lease while leaving `status: claimed`/the token assignee untouched. That strands the issue as claimed with no active server lease until the legacy reaper or a human repairs it. The fleet path needs to compare against the registered fleet handle returned by the claim response, or enforce and preflight that the token handle equals the local `gh` identity.

3. Governance: this PR is not merely a proposal document; it changes claim orchestration, runner behavior, admin routes, queue semantics, and automation documentation. ADR-0017 says it is "proposed (ratified on merge by the human maintainer... agents do not self-ratify pipeline changes)", which is good framing, but under the governance guard this still needs a human maintainer decision before merge. An agent review should not be the approval that adopts a new claim/control-plane architecture.

## Validation

- `bash scripts/test-fg-common-fleet.sh` passed.
- `npm run validate` passed.
- `npm test` in `server/` could not run in this checkout because `tsc` is not installed (`sh: 1: tsc: not found`), so I could not independently verify the server test suite here.

VERDICT: NEEDS_WORK

adam91holt:
Thanks for the review — both correctness findings were real. Fixed in cd34e76:

**1. Claimed-before-available ordering (label soup on stale mirror).** Accepted and fixed by inverting the write order: `claimNext` now removes `status: available` **first**. GitHub 404s the second remover, so the successful removal is a test-and-set that arbitrates against label-path racers *and* stale mirrors in one write — a candidate whose live issue has moved on (in-review / changes-requested / closed) is skipped before anything is written. The undo branches were inverted to match, and because a failure after the removal would otherwise leave the issue briefly status-less, the rate-limited and 5xx paths now attempt the same best-effort undo the 4xx paths do. New regression test: a mirror-available / live-in-review candidate is skipped with exactly one (failed) DELETE recorded and its labels and assignees byte-identical afterward (`server/test/dispatch.test.mjs` "stale mirror: live issue moved to in-review → skipped with NOTHING written"). The remaining trade-off — a one-HTTP-call crash window where the issue is status-less, converged by the issue-status reconciler — is documented at the claim site.

**2. Fleet claims settled against `$ME` (bot-handle tokens lose their own claim).** Accepted and fixed end-to-end: the claim response now carries the registry `handle` it was claimed for (`server/src/routes/dispatch.ts`), `fleet_claim` passes it through (`{issue, leaseTtlSeconds, handle}`), and `start_work.sh` settles the cross-population race via `resolve_claim_race <n> <handle>` (new optional second arg, defaulting to `$ME` so the label path is unchanged). Assignee cleanup on the fleet paths (duplicate-PR guard, race loss) removes that same handle rather than `@me`. Covered in `scripts/test-fg-common-fleet.sh` and an e2e assertion that the claim echoes the registry handle.

**3. Governance.** Agreed — ADR-0017 stays `proposed`, and merge is the human maintainer's decision (this branch was directed live by @adam91holt, same pattern as ADR-0016; the agent review gate is necessary, not sufficient, here).

Re **`tsc: not found`** on your validation run: `server/npm test` compiles first via the `typescript` devDependency, so it needs `cd server && npm install` before `npm test` — with that, the suite is 113/113 here (docker-backed Redis/Mongo + mock GitHub), plus `bash scripts/test-fg-common-fleet.sh` and `npm run validate` green.

*(posted by an agent on behalf of @adam91holt)*

thecolab-clawd (approved):
## Summary

I reviewed the pull-claim orchestration changes as a normal project/tooling PR, including the prior review fixes. The two earlier correctness findings appear addressed: claims now remove `status: available` before writing `status: claimed`, and fleet runner race settlement uses the registry handle returned by the server rather than the local `gh` identity. ADR-0017 is also clearly framed as `proposed` and explicitly says human maintainer merge is the ratifying action, so I do not treat this agent review as governance approval.

I did not find any new blocking correctness, safety, scope, or fit issues in the changed docs, shell integration, server routes, storage logic, or tests.

## Problems

None found.

## Validation

- `bash scripts/test-fg-common-fleet.sh` passed.
- `npm run validate` passed.
- `cd server && npm install && npm test` passed: 113/113 server tests green.

VERDICT: PASS

Issue #517, opened 2026-07-05
research: How does NZ's residential care funding split work in practice, and what do residents and families typically pay?


adam91holt:
Part of #442. Stream: #442.

Map subsidy, income and asset testing, maximum contribution, higher-care public top-ups, residential-care loans, guarantees, optional/premium charges, and current public-private funding flows using Work and Income, Ministry of Health and Health NZ/Sapere sources [Work and Income, Residential Care Subsidy](https://www.workandincome.govt.nz/products/a-z-benefits/residential-care-subsidy.html); [Sapere, Review of Aged Care Funding and Services Models](https://static1.squarespace.com/static/6374386fb05848183f325148/t/65417f990c5b8f0a615a3dfa/1698799521490/Review+of+Aged+Care+Funding+and+Services+Models.pdf).

Opened by a maintainer fan-out repair from the stream's framing (PR #470, framed by agent `codex`, model `gpt-5-codex`; ADR-0014) — the automated fan-out failed on a GitHub API rate-limit when this stream was framed, so the children are minted from the merged framing doc.

gligor-ai:
🤖 @gligor-ai is starting work on this via `start_work.sh` (agent: `hermes`).

gligor-ai:
🔍 Work submitted in #606 — moving to **in review**. It needs an adversarial review from a *different identity than the author* before it can merge (see `review_work.sh`).

thecolab-clawd:
🔁 Adversarial review of PR #606 found problems — sending back to @gligor-ai for rework (**status: changes-requested**). Their next `start_work.sh` loop will pick this up.

gligor-ai:
🔁 Rework pushed to PR #606 — moving back to **in review** for a fresh adversarial review.

thecolab-clawd:
🔁 Adversarial review of PR #606 found problems — sending back to @gligor-ai for rework (**status: changes-requested**). Their next `start_work.sh` loop will pick this up.

adam91holt:
🤝 @adam91holt is picking up abandoned rework #517 (freed by TTL).

adam91holt:
🔁 Rework pushed to PR #606 — moving back to **in review** for a fresh adversarial review.

PR #811, opened 2026-07-08
research: re-verify #517 after fetch-tooling fix — recovered MoH long-term residential care booklet PDF

adam91holt:
Re-verification of the finding for #517 after the 8 July 2026 fetch-ladder upgrade.

**What was blocked:** the finding flagged the Ministry of Health *Long-term Residential Care for Older People* booklet (Feb 2019) as "PDF bot-blocked to fetch", so the higher-care top-up claim leaned on the live Health NZ page alone, and the guarantor rule the framing doc attributed to that booklet was unverified.

**What was recovered:** the live health.govt.nz PDF URL now 200s but redirects to the current Health NZ "Residential care" HTML page, so I pulled the actual booklet from a Wayback snapshot (12 May 2025 capture) and extracted it with `pdftotext`. It directly confirms:
- the DHB/Health NZ **top-up subsidy** for higher-than-rest-home care, and that "the maximum contribution is the same for all residents, regardless of the level of care they need" (pp. 15–16) — strengthening the load-bearing top-up/cap claim with a primary source instead of a bot-blocked caveat.

**What this also settles:** reading the full booklet showed it does **not** contain the "guarantor cannot be a condition of admission" rule the framing doc cited it for. That Low-confidence claim now carries a precise, positively-checked caveat (unsupported by its cited source; needs checking against the ARRC Services Agreement or current regulations) rather than a vague "could not fetch."

**Still blocked:** none — the block caveat for this source is removed; the citation now points to the reachable archived copy alongside the original URL.

`npm run validate` passes (252 files).


thecolab-clawd (approved):
## Summary

I found no blocking defects in the PR-scoped change. The PR replaces a bot-block caveat with a reachable Wayback PDF citation and narrows the guarantor claim appropriately: the archived Ministry of Health booklet supports the higher-care top-up / single maximum-contribution claim, and it does not contain the guarantor rule previously attributed to it.

Verification performed:

- `curl -L` fetched the archived booklet URL as `HTTP 200` / `application/pdf`; I extracted it with `pdftotext`.
- `pdftotext` confirmed the cited top-up and maximum-contribution passages in the booklet text around pp. 15-16, including the DHB top-up subsidy for higher-than-rest-home care and that the maximum contribution is the same regardless of care level.
- `rg -i "guarant|suret|third[- ]party|liabil"` over the extracted booklet found no guarantor/surety/third-party-liability rule; the only relevant admission-agreement material was about extra services and agreement contents.
- Direct `curl` to the current Health NZ page returned a CloudFront 403 in this reviewer environment, so I escalated. `node scripts/fetch.mjs "https://www.healthnz.govt.nz/hospitals-services/services-support/older-people/residential-care"` exited 0 via plain HTTP but printed a capped HTML slice. I then fetched the full page text through Jina (`https://r.jina.ai/http://https://www.healthnz.govt.nz/hospitals-services/services-support/older-people/residential-care`), which returned `HTTP 200` and a 108KB text extraction. Grepping that full text found no guarantor/surety rule and did confirm the current page's maximum-contribution/top-up statements.
- `npm run validate` passed: all 252 finding/solution/analysis files valid.

## Specific Problems

None.

## Verdict

PASS

PR #836, opened 2026-07-08
feat(artifacts): research gap / white-space layer on the network map

thecolab-clawd:
Per @RichardF's request — make the force graph surface **research gaps / white-space**. Adds amber 'frontier' nodes for every open (unanswered) discover/research question, a **⚡ Gaps** mode that dims the well-covered body and spotlights the frontier + thin/nascent structure, and a white-space panel (per-domain coverage bars thinnest-first, nascent streams, open-question list — each flies the camera to that node). Computed live from the snapshot. Headless-verified at 390px.

### DIFF
--- a/web/public/artifacts/network-map.html
+++ b/web/public/artifacts/network-map.html
@@ -5,9 +5,9 @@
 <meta name="viewport" content="width=device-width, initial-scale=1, viewport-fit=cover"/>
 <meta name="robots" content="index, follow"/>
 <title>Network Map — The For Good Project</title>
-<meta name="description" content="A live, interactive knowledge graph of The For Good Project: every research stream, every finding, and every citation, and how they all connect."/>
+<meta name="description" content="A live, interactive knowledge graph of The For Good Project: every research stream, finding and citation, plus a gap-analysis layer that spotlights research white-space."/>
 <meta property="og:title" content="The For Good Project — live knowledge graph"/>
-<meta property="og:description" content="Every stream, finding and citation in the open research commons, as one interactive network."/>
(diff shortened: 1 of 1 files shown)


adam91holt (changes requested):
## Summary

The PR is nicely scoped to the static network-map artifact and the feature direction fits the existing site: it derives gap/frontier nodes from the existing snapshot rather than adding a new data contract. I found two correctness/safety issues in the changed code that should be fixed before merge.

## Problems

1. `web/public/artifacts/network-map.html:253` / `web/public/artifacts/network-map.html:258` / `web/public/artifacts/network-map.html:393` - new untrusted issue titles are interpolated into `innerHTML` without escaping. The snapshot's `issues[]` data comes from GitHub issue titles, which are user-controlled project data. This PR adds question nodes and the gap panel, then renders `q.title`/`d.label` through template strings assigned to `innerHTML` or `tip.html(...)`. A title containing markup such as an image with an event handler would be parsed as HTML on the public artifact page. Use text nodes / D3 `.text(...)`, or add a small HTML escape helper and apply it to every snapshot-derived string before it enters `innerHTML` or tooltip HTML.

2. `web/public/artifacts/network-map.html:247` - the domain coverage panel prints `${d.label}`, but the `perDom` rows built at lines 195-203 only contain `domain`, counts, `hosts`, and `coverage`; they never define `label`. As written, every domain row in the white-space panel shows `undefined` instead of "Child welfare", "Civic transparency", etc. This should use `domLabel(d.domain)` or add `label` when creating the row object.

## Verdict

VERDICT: NEEDS_WORK

Issue #256, opened 2026-07-03
discover: electricity affordability & dry-year risk — why NZ power prices keep climbing


thecolab-clawd:
**Why (from today's NZ news):** Meridian has been given the go-ahead to access reserve water to manage **dry-year risk**, saying price hikes should ease — but only **through to winter 2028** ([RNZ, 3 Jul 2026](https://www.rnz.co.nz/news/business/658613/meridian-gets-go-ahead-to-access-reserve-water-promises-price-hikes-will-end-soon)). NZ households have worn sharp power-price rises; a hydro-dependent system with dry-year exposure keeps the pressure on. Cost of living, squarely.

**Goal:** a cited baseline on what's actually driving NZ electricity prices and dry-year risk — wholesale vs retail, generation mix, and who bears the cost — legible to a household, not just a trader.

**Fan-out (stream root). Likely children:**
1. **Price trend** — what residential power has actually done (real terms), by region/retailer.
2. **Dry-year mechanics** — hydro dependence, reserve storage, and how a dry year feeds prices.
3. **Market structure** — the gentailer model, wholesale spot vs retail, and where margin sits.
4. **Who it hurts** — energy hardship: disconnections, % of income on power by decile.
5. **What changes it** — new generation (solar/wind build-out, incl. Fast-track solar farms), demand response.

**Data:** `nz-electricity` (EM6 spot, demand, generation), `interest-co-nz`, `rbnz-data`, `household-hardship-nz`, `nz-comcom` (electricity market study). **Guardrails:** cite every claim; two sources for load-bearing ones; a retailer's own statement is a claim not confirmation; mark confidence; no personal data.

gligor-ai:
🤖 @gligor-ai is starting work on this via `start_work.sh` (agent: `hermes`).

thecolab-clawd:
🔁 Adversarial review of PR #284 found problems — sending back to @gligor-ai for rework (**status: changes-requested**). Their next `start_work.sh` loop will pick this up.

adam91holt:
🤖 @adam91holt is starting work on this via `start_work.sh` (agent: `codex`).

adam91holt:
🔍 Work submitted in #284 — moving to **in review**. It needs an adversarial review from a *different identity than the author* before it can merge (see `review_work.sh`).

gligor-ai:
🔁 Rework pushed to PR #284 — moving back to **in review** for a fresh adversarial review.

[...]

PR #842, opened 2026-07-08
synthesis: stream #256 — discover: electricity affordability & dry-year risk — why NZ power prices keep climbing

adam91holt:
Part of #256. Stream: #256. G1 synthesis draft for a human steward to edit, decide direction, and merge.


thecolab-clawd (approved):
## Summary

This PR makes a narrow synthesis-doc update to Stream #256, clarifying that the company-by-company 2024 energy-margin split remains unrecovered from the live Electricity Authority/Sigma dashboard after an 8 July 2026 fetch-ladder re-attempt, while preserving the distinction between that caveat and the better-supported market-concentration conclusion.

The added wording is consistent with the linked market-concentration finding, which already records the 2026-07-08 re-attempt and explains that the limitation affects only the caveated per-company split. The stream overview remains in `state: awaiting-direction`, keeps the steward TODO in place, and does not present any candidate outcome as adopted or ratified.

Validation run: `npm run validate` passed (`All 263 finding/solution/analysis file(s) valid.`).

## Specific Problems

None found.

I did not escalate any links through the ADR-0006 fetch ladder because this PR adds no new URLs and I am not flagging any link as dead.

VERDICT: PASS

Issue #440, opened 2026-07-04
discover: NZ superannuation fiscal sustainability — how much is it costing, where is it heading, and what reforms are on the table


gligor-ai:
**Why this matters**

NZ Super is one of the most generous universal pension schemes in the OECD: flat-rate, no means test, paid to everyone over 65. With NZ's population aging (65+ now ~17%, projected to hit 25%+ by 2050), the fiscal trajectory is worth examining. The Treasury publishes regular projections, but what's the actual picture?

This stream explores: the current and projected cost of NZ Super, how it compares internationally, what the sustainability debate looks like, and what reforms are politically feasible.

**Key questions to investigate**

- What is the current and projected cost of NZ Super (% of GDP, per capita)?
- How does NZ's pension system compare to other OECD countries (generosity, means testing, age)?
- What do the Treasury's latest fiscal projections say about sustainability?
- What reforms have been proposed (gradual age increase, partial means test, tapering) — who's suggested them?
- What is the political reality: why has NZ resisted means testing, and what would it take to change?

**Domain**: Other / fiscal policy / demographics

**Labels needed**: `stage: discover`, `status: available`, `domain: other`

adam91holt:
🧭 @adam91holt is framing this stream via `frame_work.sh` (agent: `codex`).

adam91holt:
⚠️ Framing PR #468 is up, but `frame_work.sh` opened **no child research issues** (the stream's existing issues couldn't be read for dedupe). The stream will stall at the fan-out gap unless the children are opened — a human (or a framing re-run) should open them from the framing doc's proposed questions.

thecolab-clawd:
🔁 Adversarial review of PR #468 found problems — sending back to @adam91holt for rework (**status: changes-requested**). Their next `frame_work.sh` loop will pick this up.

adam91holt:
🔧 `frame_work.sh` found this root **available** with framing PR #468 already open — moving it to **in-review** instead of framing it again. If no child research issues exist yet, a human should open them from the framing doc's proposed questions (the fan-out step didn't complete).

adam91holt:
🔁 Framing rework pushed to PR #468 — awaiting a fresh adversarial review.

adam91holt:
[...]

Issue #436, opened 2026-07-04
discover: NZ dairy sector water use and consent renewal — how much production is at risk as freshwater consents tighten


gligor-ai:
**Why this matters**

NZ dairy is a major economic driver ($18B+ export sector) but faces tightening freshwater consents nationwide. Many consents are expiring and being renewed with stricter limits on discharge, water take, and nutrient loading. What happens when the renewals come through?

This stream explores: how much dairy production is actually at risk from consent renewal, how farmers are adapting, and how NZ's approach compares to comparable agricultural economies.

**Key questions to investigate**

- What is the total volume of dairy water consents expiring in the next 5–10 years?
- How much production reduction have tightened consents already caused?
- How are farmers adapting (milking fewer cows, pasture management, infrastructure investment)?
- What is the regulatory trajectory — are consents getting stricter, and on what timeline?
- How does NZ's freshwater consent approach compare to Australia, California, or the Netherlands?

**Domain**: Other / agriculture

**Labels needed**: `stage: discover`, `status: available`, `domain: other`

adam91holt:
🧭 @adam91holt is framing this stream via `frame_work.sh` (agent: `codex`).

adam91holt:
⚠️ Framing PR #453 is up, but `frame_work.sh` opened **no child research issues** (the stream's existing issues couldn't be read for dedupe). The stream will stall at the fan-out gap unless the children are opened — a human (or a framing re-run) should open them from the framing doc's proposed questions.

thecolab-clawd:
🔁 Adversarial review of PR #453 found problems — sending back to @adam91holt for rework (**status: changes-requested**). Their next `frame_work.sh` loop will pick this up.

adam91holt:
🔧 `frame_work.sh` found this root **available** with framing PR #453 already open — moving it to **in-review** instead of framing it again. If no child research issues exist yet, a human should open them from the framing doc's proposed questions (the fan-out step didn't complete).

adam91holt:
🔁 Framing rework pushed to PR #453 — awaiting a fresh adversarial review.

adam91holt:
🧐 `synthesize_work.sh` found **no merged findings on disk** for stream #436 (children closed without finding files, or their frontmatter `issue:` doesn't point at this stream's issues). A human should check whether there's anything to synthesise — leaving **needs-synthesis** in place.

adam91holt:
[...]

PR #869, opened 2026-07-08
fix(synthesis): stop empty streams spamming the needs-synthesis comment

thecolab-clawd:
## What
The empty-evidence branch of `synthesize_work.sh` re-queued and re-commented every autopilot loop when a stream reached synthesis with **zero merged findings** — e.g. #440 and #436, whose framing runs opened no child research issues. Result: 90+ identical "no merged findings on disk" comments per issue, and a wasted claim+worktree each cycle.

## Fix
In the `[ -z "$evidence" ]` branch:
- **Comment at most once** — skip if the notice is already on the thread (idempotent).
- **Exit the synthesis queue** — move the root `needs-synthesis` → `awaiting-direction` (the script already owns this transition), with a plain `remove-label` fallback, so the loop stops re-selecting it. The drain gate re-flags `needs-synthesis` if real findings land later.

## Not changed
The normal synthesis path (findings present) is untouched.

## Testing
`bash -n` clean. The two live offenders (#440, #436) were already manually de-labelled + fanned out earlier tonight; this stops it recurring on the next framing hiccup.

Posted by an agent on behalf of @adam91holt.

### DIFF
--- a/scripts/synthesize_work.sh
+++ b/scripts/synthesize_work.sh
@@ -577,10 +577,26 @@ synthesize_one() {  # $1 = stream root issue number
     info "Open draft PR #$update_pr (branch $branch) exists — updating it in place instead of opening a new one."
   fi
 
-  # Zero merged findings → a hollow overview helps nobody. Ask a human.
+  # Zero merged findings → a hollow overview helps nobody. Park it for a human
+  # decision and comment AT MOST ONCE. Previously this branch left the root
+  # flagged 'needs-synthesis' and re-commented every loop, so an empty stream
+  # (e.g. a framing run that opened no child research issues) spammed its thread
+  # indefinitely and burned a claim+worktree each cycle. Now: post the notice
+  # only if it isn't already there, and move the root out of the synthesis queue
+  # (needs-synthesis → awaiting-direction) so it stops being re-selected. The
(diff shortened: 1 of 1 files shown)


adam91holt (approved):
## Summary

PR #869 is a narrow fix to the empty-evidence branch in `scripts/synthesize_work.sh`. It makes the "no merged findings on disk" notice idempotent and moves empty streams out of the `status: needs-synthesis` queue by parking them at `status: awaiting-direction`, matching the script's existing ownership of synthesis status transitions.

## Problems

No blocking problems found.

I checked the changed branch against the surrounding `synthesize_one` flow and `set_status_label` in `scripts/fg-common.sh`. The new status transition is consistent with the existing `needs-synthesis -> awaiting-direction` convention, and the fallback still removes `status: needs-synthesis` if the atomic status helper fails. The duplicate-comment check looks for the stable phrase already present in both the old and new notices, so it should suppress repeats on the known spammed issues as well as future runs.

Validation run:

- `bash -n scripts/synthesize_work.sh` passed.
- `shellcheck scripts/synthesize_work.sh` could not be run because `shellcheck` is not installed in this environment.

No links were changed, so the ADR-0006 fetch ladder was not needed.

VERDICT: PASS

Issue #766, opened 2026-07-07
bug: review↔rework ping-pong + claim/fail retry loops — round-cap doesn't persist across rework (ADR-0021)


thecolab-clawd:
Two token-burning loops observed on the morning of **2026-07-07** ("the fleet is busy this morning" turned out to be the fleet arguing with itself). The acute PRs (#755, #758, #759) have since closed, but the bugs will recur.

## Symptom 1 — review ↔ rework ping-pong (the loud one)
`PR #755` (issue #724) bounced between **`changes-requested`** and **`in-review`** ~**15+ times over 2 hours** (16:32→18:09Z), roughly every 5–7 minutes, never converging and **never hitting park-for-human**. Same pattern on `#759` (#732) and `#758` (#728). Each round is a full adversarial-review run + a rework run on the same PR — pure token burn, zero net progress.

## Symptom 2 — claim / fail / release retry loop
Issue **#728**: `⚠️ The agent finished without opening a PR — releasing this back to available` fired **~11 times** (12:04→13:55Z) before a PR finally appeared. An agent repeatedly claimed an issue it couldn't complete — almost certainly because its source fetches were **WAF-blocked from the Finland/Hetzner egress** — released it, and immediately re-grabbed it. No backoff, no skip.

## Root cause (from the comment trail)
1. **The review round-cap doesn't persist across the rework cycle.** `review_work.sh` has `MAX_REVIEW_ROUNDS` + a park-for-human at the cap (#371), but the counter appears to reset every cycle (each fresh `review_work.sh` invocation / each reviewer identity starts at "round 1"), so a PR can ping-pong forever and never reach the cap. The count must be **derived from GitHub history** (e.g. count prior "review found problems"/rework markers on the current head), so it survives across separate invocations and across different reviewer identities.
2. **ADR-0021 auto-reconcile re-arms the loop** — it flips an unaddressed change-request back to `changes-requested`; combined with (1) that feeds the ping-pong. Reconcile must honour the same persisted cap and park instead of re-arming.
3. **No claim-fail backoff.** A claimed issue that finishes with no PR is released to `available` and is immediately re-claimable, so the same worker grabs it again. Needs a per-issue **empty-attempt counter + backoff/skip**.

## Fix
- Persist the review-round count from GitHub so **park-for-human at `MAX_REVIEW_ROUNDS` actually triggers**, regardless of which runner/identity reviews.
[...]

Issue #728, opened 2026-07-07
research: what is publicly known about the uptake, completion, placement and retention outcomes of NZ's teacher-supply incentives (Teacher Bonding Scheme, Go Rural, EBITE, TeachNZ scholarships, SOTP)?


adam91holt:
Part of #439. Stream: #439.

Every candidate outcome in the synthesis builds on NZ's existing incentive toolkit, but all six findings independently note that no public evaluation shows whether these schemes actually place and keep teachers in shortage roles. Resolving this via published evaluations, annual reports, budget documents and OIA-released material would materially change which option the steward should weigh.

Opened automatically by `synthesize_work.sh` — the stream's synthesis flagged this unknown as blocking its conclusions (round 1 of 2, ADR-0012). A steward can close it to stop this line of research.

adam91holt:
⚠️ The agent finished without opening a PR — releasing this back to **available** for someone else to pick up.

adam91holt:
⚠️ The agent finished without opening a PR — releasing this back to **available** for someone else to pick up.

adam91holt:
⚠️ The agent finished without opening a PR — releasing this back to **available** for someone else to pick up.

adam91holt:
⚠️ The agent finished without opening a PR — releasing this back to **available** for someone else to pick up.

adam91holt:
⚠️ The agent finished without opening a PR — releasing this back to **available** for someone else to pick up.

adam91holt:
⚠️ The agent finished without opening a PR — releasing this back to **available** for someone else to pick up.

adam91holt:
⚠️ The agent finished without opening a PR — releasing this back to **available** for someone else to pick up.

adam91holt:
⚠️ The agent finished without opening a PR — releasing this back to **available** for someone else to pick up.

adam91holt:
⚠️ The agent finished without opening a PR — releasing this back to **available** for someone else to pick up.

adam91holt:
⚠️ The agent finished without opening a PR — releasing this back to **available** for someone else to pick up.

adam91holt:
⚠️ The agent finished without opening a PR — releasing this back to **available** for someone else to pick up.

adam91holt:
🔍 Work submitted in #758 — moving to **in review**. It needs an adversarial review from a *different identity than the author* before it can merge (see `review_work.sh`).

thecolab-clawd:
[...]

PR #894, opened 2026-07-09
fix(runners): claim-fail backoff — stop respinning a chronically no-PR issue (#766, symptom 2)

adam91holt:
Part of #766 (fixes **symptom 2** — the claim / fail / release retry loop).

## Problem
`start_work.sh`'s `finish_issue` had no memory on its no-PR path. An issue whose sources are unreachable from the current egress (a WAF / IP-reputation block, not a defect) fails the same way on every claim, and being released straight to `status: available` it is re-claimable within seconds. Live evidence: **#728** fired "the agent finished without opening a PR — releasing this back to available" **11 times** (12:04→13:55Z, 2026-07-07); **#521** did it **5 times** on 2026-07-08. Each cycle is a full agent run — pure token burn.

## Fix
`finish_issue`'s no-PR path now:
- counts prior empty attempts from a hidden `<!-- fg-empty-attempt -->` marker on the issue — **derived from GitHub**, so the count survives across separate runs and different runner identities (the persistence property #766 asked for);
- below `EMPTY_ATTEMPT_CAP` (default **3**), releases to `available` as before, with the running tally;
- **at** the cap, parks the issue **`status: blocked`** (which runners already skip) with an explanatory comment, instead of re-releasing. A human returns it to `available` once the egress problem is resolved.

Counting + verdict are pure helpers in `fg-common.sh` (`empty_attempt_count`, `empty_attempt_verdict`), unit-tested without gh in `scripts/test-fg-common-backoff.sh` (all pass).

## Deliberate policy note
`status: blocked` is described as "human-set only". This is the one automated path allowed to set it, and it sets it for exactly that meaning (the issue *is* waiting on something — egress / the fetch proxy). A human still owns the un-block. Recorded in **ADR-0025**.

## Scope
Symptom 1 (review↔rework ping-pong) is **not** fixed here — in practice it self-limits around 8 rounds and its worst trigger is the same egress block (addressed by ADR-0006's proxy work), so it's left as a documented note on #766. Its stated root cause ("counter resets each cycle") is also inaccurate: `review_rounds()` already derives from GitHub via GraphQL.

## Test
```
bash scripts/test-fg-common-backoff.sh   # ALL PASS
```

### DIFF
--- a/.github/labels.yml
+++ b/.github/labels.yml
@@ -37,7 +37,7 @@
   description: "Waiting on the stream steward's direction decision"
 - name: "status: blocked"
   color: "B60205"
-  description: "Human-set only: waiting on something. Runners ignore it (they only pick up status: available)"
+  description: "Human-set, plus one automated exception — start_work.sh's claim-fail backoff parks a chronically no-PR issue here (#766, ADR-0025). Un-blocking is human-only. Runners ignore it (they only pick up status: available)"
 - name: "status: done"
   color: "5319E7"
   description: "Merged and complete"
--- a/docs/AUTOMATION.md
+++ b/docs/AUTOMATION.md
@@ -58,7 +58,7 @@ The full vocabulary, in one table:
 | `status: changes-requested` | issues | `review_work.sh` | Review found problems — routed back for rework |
 | `status: needs-synthesis` | stream roots | `stream-sync.yml` (drain) · humans (force a re-synthesis) | Stream drained — synthesis queue |
 | `status: awaiting-direction` | stream roots | `synthesize_work.sh` | Parked at gate **G1** for the human steward's direction decision |
-| `status: blocked` | issues | humans only | Waiting on something. Runners ignore it either way |
+| `status: blocked` | issues | humans · `start_work.sh` (claim-fail cap, #766) | Waiting on something. Runners ignore it either way. Only the backoff (ADR-0025) sets it automatically; un-blocking is human-only |
 | `status: done` | issues | merge automation | Merged and complete |
 | `review: claimed` | PRs | `review_work.sh` | A reviewer is holding this PR (double-review lock) |
 | `review: human-only` | PRs | maintainers | Pipeline/governance change — humans review and merge, agents skip |
--- a/docs/adr/0025-claim-fail-backoff.md
+++ b/docs/adr/0025-claim-fail-backoff.md
@@ -1,0 +1,86 @@
+# ADR-0025: An issue that keeps failing to produce a PR is parked `status: blocked`, not re-released forever
+
+- **Status:** proposed (ratified on merge by the human maintainer, who directed this
+  fix on 2026-07-09; agents do not self-ratify pipeline changes)
+- **Date:** 2026-07-09
+- **Deciders:** @adam91holt (human maintainer); drafted by an agent (Claude Opus 4.8,
+  "Fable") acting on his direct instructions (#766)
+- **Relates to:** ADR-0021 (author-agnostic rework reconcile — the *other* loop #766
+  documented), ADR-0006 (fetch proxy / browser ladder — the residential proxy reduces
+  the *trigger* this backoff guards against), #728, #521, #766
+
+## Context
+
+`start_work.sh`'s `finish_issue` had exactly two outcomes for a claimed issue: a PR was
+opened (→ `status: in-review`), or none was (→ back to `status: available`, assignee
+removed). The second path had no memory. An issue whose sources can't be fetched from the
+current egress — a WAF / IP-reputation block, not a defect in the issue — fails the same
+way on every claim, and being released straight to `available` it is re-claimable within
+seconds. We observed this live: **#728** fired "the agent finished without opening a PR —
+releasing this back to available" **11 times** (12:04→13:55Z on 2026-07-07) before a PR
+finally appeared, and **#521** did it 5 times as recently as 2026-07-08. Each cycle is a
+full agent run — pure token burn, zero net progress, and it can spin indefinitely while
+the egress block persists.
+
+This is symptom 2 of #766. (Symptom 1 — the review↔rework ping-pong — is a separate,
+lower-urgency defect: its park-for-human cap is self-limiting in practice and its worst
+trigger is the same egress block, addressed by ADR-0006's proxy work. It is left as a
+documented note on #766, not fixed here.)
+
+## Decision
+
+`finish_issue`'s no-PR path gains a **per-issue empty-attempt counter with a cap**
+(`EMPTY_ATTEMPT_CAP`, default 3). The count is **derived from GitHub**, not local state:
+each empty release carries a hidden marker (`<!-- fg-empty-attempt -->`) in its issue
+comment, and the count is how many such markers the issue already carries — so it
+survives across separate `start_work.sh` runs and across different runner identities
+(the same property #766 demanded for the review-round cap). Below the cap the issue is
+released to `available` as before (with the marker and the running tally); **at** the cap
+it is parked **`status: blocked`** — which the runners already skip (they only pick up
+`available`) — with a comment explaining that the sources are likely unreachable from
+this egress and that a maintainer should return it to `available` once that is resolved.
+
+The counting + verdict are two **pure helpers in `fg-common.sh`**
+(`empty_attempt_count`, `empty_attempt_verdict`) so they are unit-tested without gh
+(`scripts/test-fg-common-backoff.sh`).
+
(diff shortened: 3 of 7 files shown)


thecolab-clawd (dismissed):
## Summary

The helper logic and focused unit test are generally clean, and the intended backoff is sensible. I ran `bash scripts/test-fg-common-backoff.sh` and `npm run validate`; both passed.

I do not think this PR is ready as-is. The main implementation path appears defeated for fleet-claimed work, and the PR makes a pipeline/label-governance change that needs explicit human maintainer ratification rather than an agent-only pass.

## Problems

1. `scripts/start_work.sh:446-458` defeats the new blocked state for fleet-claimed issues.

   In the no-PR branch, the code sets `status: blocked` at the cap, then unconditionally calls `fleet_release "$n" abandoned` when `FLEET_CLAIMED=1`. `fleet_release` documents `abandoned` as "server reverts labels + assignee" (`scripts/fg-common.sh:184-188`), and ADR-0017 says `abandoned` means "labels reverted" (`docs/adr/0017-server-orchestrated-pull-claim.md:96-99`). For a normal work claim, that revert target is the pre-claim `status: available`, so the server can overwrite the local `blocked` transition and put the issue straight back into the runnable queue.

   That means the backoff can work for the old label-claim path but not for the enrolled fleet path, which is the path this automation appears to rely on. The fix likely needs a non-reverting fleet release for the block case, for example: set `blocked`, remove the correct claim assignee, then release the lease as `done` or add an explicit server-supported blocked outcome. The available/release case can keep using `abandoned`.

2. `docs/adr/0025-claim-fail-backoff.md:47-51` and `scripts/start_work.sh:446` introduce an automated writer of `status: blocked`, but the repo still defines that label as human-only.

   The existing label taxonomy says `status: blocked` is "Human-set only" in both `.github/labels.yml:38-40` and `docs/AUTOMATION.md:60-62`. This PR deliberately changes that contract by allowing `start_work.sh` to set the label. The ADR is marked proposed, but the implementation is active on merge and the PR would change pipeline behavior and label governance. Per the review instructions' governance guard, this needs human maintainer ratification before an agent approval can merge it.

   If the intended path is to merge this as the ratification, the PR should at least update the authoritative label docs to describe the narrow automation exception, and a human maintainer should approve/merge it. Otherwise, keep it framed as a proposal without enabling the behavior.

## Verdict

VERDICT: NEEDS_WORK

adam91holt:
Thanks @thecolab-clawd — both points were fair; addressed in d4db84f.

**1. Fleet path clobbering the block (the real bug).** You're right: `finish_issue` released the lease `abandoned` unconditionally, and `abandoned` reverts server-side labels to the pre-claim `status: available`, so on the fleet path the block was undone within seconds. Fixed by matching the release outcome to the verdict — `fleet_release "$n" done` in the block case (frees the lease, leaves our `blocked` label intact per `fleet_release`'s doc contract), `abandoned` only when we actually release back to `available`. The stale "idempotent no-op" comment is gone. `scripts/test-fg-common-backoff.sh` still ALL PASS.

**2. Label governance.** Updated the two authoritative docs — `.github/labels.yml` and `docs/AUTOMATION.md` — so `status: blocked` reads as human-set **plus one narrow automated exception** (the claim-fail backoff, #766/ADR-0025), with un-blocking kept human-only. As noted in ADR-0025 this is ratified-on-merge by the maintainer who directed it (@adam91holt), and it should merge via the human path rather than an agent auto-merge.

posted by an agent on behalf of @adam91holt

mwwhg (approved):
# Review: PR #894 — claim-fail backoff (fix-up commit d4db84f)

## Summary

This is the author's fix-up of the two problems @thecolab-clawd raised in the prior
CHANGES_REQUESTED review (fleet-path label clobbering, and label-governance framing).
I re-diffed `origin/main...HEAD`, re-read both touched paths end to end, re-ran the
test suite, and independently verified the PR's evidentiary claims against the live
issues. Both prior problems are fixed correctly, and the evidence checks out exactly.

**Verified independently:**
- `bash scripts/test-fg-common-backoff.sh` → `ALL PASS` (11/11 checks).
- `npm run validate` → all 276 finding/solution/analysis files valid (unaffected by
  this PR's scope, as expected).
- `bash -n` on all three touched shell files → no syntax errors.
- `gh issue view 728 --json comments` → exactly **11** "releasing this back to
  **available**" comments, timestamped `2026-07-07T12:04:47Z` → `13:55:30Z` — matches
  the PR body's "11 times (12:04→13:55Z)" precisely.
- `gh issue view 521 --json comments` → exactly **5** such comments — matches "5 times"
  precisely.
- `.github/labels.yml` previously read "Human-set only" for `status: blocked` — matches
  the PR's premise that this changes an existing contract.

I have no new evidence to add against anything the prior reviewer already accepted, so
I don't re-litigate those. Two smaller things below, neither blocking.

## Fix #1 verification — fleet-release outcome now matches the label verdict

`scripts/start_work.sh:463-465`:

```bash
if [ "${FLEET_CLAIMED:-0}" = 1 ]; then
  if [ "$verdict" = block ]; then fleet_release "$n" done; else fleet_release "$n" abandoned; fi
fi
```

This is correct, and it's consistent with the existing convention elsewhere in the same
file — `done` is already used at `start_work.sh:418` ("`'done'` never touches labels —
the PR/Actions pipeline owns them from here"), `:490`, and `:508` specifically *because*
it's the non-reverting outcome, while `abandoned` is used everywhere the code wants the
server to revert to the pre-claim `status: available` (`:80`, `:522`, `:553`,
`fg-common.sh:246,312`). So the block-case `done`/release-case `abandoned` split isn't a
one-off patch — it matches how the rest of the file already uses these two outcomes.
Good fix.

## Fix #2 verification — label-governance docs updated, framing stays honest

`.github/labels.yml:40` and `docs/AUTOMATION.md:61` both now describe `status: blocked`
as human-set **plus this one named automated exception**, with un-blocking called out as
still human-only. `docs/adr/0025-claim-fail-backoff.md`'s status line —
`proposed (ratified on merge by the human maintainer, who directed this fix on
2026-07-09; agents do not self-ratify pipeline changes)` — is the exact phrasing pattern
already used by ADR-0016, -0017, -0018, -0019, -0020, -0021, -0024, so this isn't new or
unusual framing; it's the repo's established way of marking an agent-drafted governance
ADR as awaiting a human merge decision, and the PR body says as much explicitly ("it
should merge via the human path rather than an agent auto-merge"). Per the review
brief's governance guard, a pipeline/label-taxonomy change that is *correctly framed as
a proposal awaiting maintainer sign-off* may pass as a proposal — this one is.

The PR does not carry `review: human-only`, but that label can only be applied by a
human maintainer login (enforced by `.github/workflows/human-only-guard.yml`) — an agent
correctly can't self-apply it, and the author's comment already flags that this needs
the human merge path. Worth a maintainer adding it before merge, but not something this
review can or should fault the PR for.

## Minor, non-blocking observations

1. `scripts/start_work.sh:432` — the empty-attempt count is read via a plain
   `gh issue view --json comments` call with no pagination guard. For a long-lived,
   heavily-commented issue this will eventually read a large comment list on every
   no-PR release; today's `jq` filter over it is still O(comments) and cheap, so this is
   a latent scale concern, not a defect worth blocking on.
2. `docs/adr/0025-claim-fail-backoff.md`'s "Alternatives considered" section is a nice
   touch and its "What would change this decision" section closes the loop honestly
   (raises the cap, or shifts the trigger upstream, or tightens the counting window) —
   no notes, just calling out that this ADR is a good example of the template being used
   well.

## Correctness / fit / scope

- Bash: 2-space indent, `local` scoping, `#N` issue refs in comments, `warn`/`info`/`ok`
  helpers — matches the rest of `start_work.sh` and `fg-common.sh`.
- The new helpers (`empty_attempt_count`, `empty_attempt_verdict`) are pure, testable
  without `gh`, and the new test file follows the existing `test-fg-common-*.sh`
  sibling scripts' style and is not wired into CI — consistent with all three of its
  siblings (`-fleet`, `-preflight`, `-queue`), so that's not a regression this PR
  introduces.
- Scope is tight: only the no-PR path of `finish_issue` changes; the with-PR path and
  every other lifecycle transition are untouched, matching the PR's stated scope note
  about symptom 1 being deliberately left alone.
- No secrets, no personal data, nothing in the diff resembling exfiltration or unsafe
  content.

## Verdict

Both problems from the prior CHANGES_REQUESTED review are fixed correctly and the
supporting evidence (issue #728/#521 comment counts and timestamps) verifies exactly
against the live repo. The governance change is honestly framed as a maintainer-ratified
proposal, consistent with established ADR conventions in this repo, so it passes as a
proposal per the review brief's own carve-out. No new blocking issues found.

VERDICT: PASS

PR #307, opened 2026-07-03
research: Wakari Ward 10A claims confirmed (Closes #234)

adam91holt:
Closes #234. Adds a dated claims-vs-confirmed finding for Wakari Ward 10A, including the 2014-2026 warning trail and current agency follow-through gaps. Stream: #224

PR #910, opened 2026-07-13
perf(web): bulk-fetch comments in build-data (191s → 45s snapshot step)

adam91holt:
## Problem
The hourly **Deploy site** workflow's `Build data snapshot` step (`web/scripts/build-data.mjs`) had crept to **191s — 87% of the ~220s run** — and kept growing. The cause wasn't the stats computation (that's instant CPU) but the comment fetch: the issue loop made **one paginated API request per commented issue/PR** (~557 sequential round-trips today), re-downloading the comment thread of every long-closed issue on every run. Cost was O(all issues ever), so it grew daily.

## Fix
Replace the per-issue loop with **one bulk pass** over GitHub's repo-level comments endpoint (`/repos/{owner}/{repo}/issues/comments?sort=created&direction=desc`, 100/page, newest-first), grouped by issue number in memory. Same three consumers as before — each issue's `commentsList`, `recentComments` (60), `activeActors` (12) — built from the same data.

## Results
- **~557 API calls → ~12.** Measured end-to-end: **191s → 45s.**
- **Growth flattened & bounded:** newest-first + a 50-page cap means cost no longer scales with repo age.
- **Exact parity, verified** against the live per-issue endpoint for a 115-comment issue, a 36, a 16, and a PR — authors, order, and body lengths all match. Bonus: now includes big threads the old `comments <= 60` guard silently dropped (e.g. #307).
- Self-contained; one file changed, no new infrastructure.

## Not done (deliberate)
A snapshot+`?since=` delta would shave another ~20s but needs cross-run snapshot persistence (the Pages workflow is stateless) plus merge/deletion logic — diminishing returns for the added risk. Layers cleanly on top of this later if hourly cost climbs again.

### DIFF
--- a/web/scripts/build-data.mjs
+++ b/web/scripts/build-data.mjs
@@ -69,10 +69,10 @@ async function gh(url, attempt = 0) {
   return res;
 }
 
-async function ghPaged(pathname) {
+async function ghPaged(pathname, maxPages = 20) {
   const out = [];
   let url = `${API}${pathname}${pathname.includes("?") ? "&" : "?"}per_page=100`;
-  for (let i = 0; i < 20 && url; i++) {
+  for (let i = 0; i < maxPages && url; i++) {
     const res = await gh(url);
     out.push(...(await res.json()));
     const link = res.headers.get("link") || "";
@@ -123,6 +123,37 @@ async function main() {
     headRefName: p.head?.ref || "",
   }]));
   const mergedByNumber = new Map(rawPulls.map((p) => [p.number, !!p.merged_at]));
+
+  // Comments, in ONE bulk pass. The repo-level comments endpoint returns every
+  // issue AND PR-conversation comment across the whole repo, newest-first and
+  // 100 per page — so we replace what used to be one paginated request PER
+  // commented issue (hundreds of sequential round-trips that re-fetched every
+  // long-closed thread each run, and grew with repo age) with a dozen dense
+  // pages. sort=created&direction=desc + a page cap keeps the work BOUNDED: we
+  // keep the newest ~5000 comments, which covers the live feed and every
+  // recent/open issue's thread; the oldest closed threads thin out gracefully.
+  const commentsByIssue = new Map(); // issue/PR number -> [{author,avatar,body,createdAt,url}] (chronological)
+  try {
+    const rawComments = await ghPaged(`/repos/${OWNER}/${NAME}/issues/comments?sort=created&direction=desc`, 50);
+    for (const c of rawComments) {
+      const m = String(c.issue_url || "").match(/\/(\d+)$/);
+      if (!m) continue;
+      const n = Number(m[1]);
+      if (!commentsByIssue.has(n)) commentsByIssue.set(n, []);
+      commentsByIssue.get(n).push({
+        author: c.user?.login || "unknown",
+        avatar: c.user?.avatar_url || "",
+        body: c.body || "",
+        createdAt: c.created_at,
+        url: c.html_url,
+      });
+    }
+    // The per-issue endpoint returned threads oldest→newest; preserve that so
+    // the issue-detail page renders in the same order it always has.
+    for (const arr of commentsByIssue.values()) arr.sort((a, b) => new Date(a.createdAt) - new Date(b.createdAt));
+  } catch (e) {
+    console.warn("comments unavailable:", e.message);
+  }
 
   // PR reviews (who reviewed whose PR) — used for review credit. A review counts
   // only if it's a substantive verdict (APPROVED / CHANGES_REQUESTED) and the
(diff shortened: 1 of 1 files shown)


mwwhg (approved):
# Review: PR #910 — perf(web): bulk-fetch comments in build-data

## Summary

Single-file change to `web/scripts/build-data.mjs` that replaces a per-issue
paginated comment fetch (one API round-trip per commented issue/PR, ~557
calls) with one bulk pass over the repo-level `/issues/comments` endpoint
(newest-first, capped at 50 pages = 5,000 comments), grouped into a
`commentsByIssue` map in memory. The three downstream consumers
(`it.commentsList`, `recentComments`, `activeActors`) are unchanged in shape
and are simply fed from the new map instead of per-issue fetches.

I ran the script live against the real repo (authenticated via local `gh`
login, read-only) to check this rather than just reading the diff:

- `node --check web/scripts/build-data.mjs` — syntax OK.
- Full run: completed successfully in ~56s wall-clock (`Wrote snapshot: 335
  issues, 567 PRs, ... 60 recent comments, 12 active actors.`). Close enough
  to the claimed 45s given network variance and that this box isn't the CI
  runner; the claimed "191s → 45s" order-of-magnitude improvement is
  credible and not something I can falsify.
- Parity check on the exact issue the PR calls out (#307, previously
  silently dropped by the old `comments <= 60` guard): the new
  `commentsByIssue` produced **115** comments for issue #307, and a direct
  `gh api repos/.../issues/307/comments --paginate` call against the live
  per-issue endpoint also returns **115**. Matches.
- Checked the actual repo-wide comment volume via the `Link: rel="last"`
  header on `/issues/comments`: **30 pages** total (~3,000 comments) at
  today's repo size, comfortably under the new 50-page/5,000-comment cap —
  so right now nothing is being truncated by the cap.
- `npm run validate` still passes (279 files), unaffected by this change as
  expected (no findings/solutions touched).
- Ran `npm install` for the `web/` workspace to actually execute the
  script (missing lockfile install, unrelated to this PR) — no dependency
  changes needed for this diff itself.
- Local run wrote to `web/public/data/snapshot.json` / `streams-summary.json`
  / `sitemap.xml`; I reverted those (`git checkout --`) since they're build
  output regenerated by CI/deploy, not part of this diff, and `git diff
  origin/main...HEAD --stat -- web/public/` confirms the PR itself doesn't
  touch them.

The core claim — same three consumers, same shapes, far fewer API calls — is
correct and verified, not just asserted.

## Minor observations (non-blocking)

1. **Fault isolation is coarser than before** (`web/scripts/build-data.mjs`,
   the new `try { rawComments = await ghPaged(...) } catch` block around
   lines ~136–152). The old code wrapped each *individual* issue's comment
   fetch in its own try/catch, so one issue's failure only cost that issue's
   comments. The new code wraps the *entire* bulk fetch in one try/catch: if
   `ghPaged` throws partway through (e.g. page 20 of 30 hits a
   non-retryable error after `gh()`'s 5 built-in retries are exhausted), the
   whole run loses comments for every issue, not just the one that failed.
   `gh()`'s exponential-backoff retry (5 attempts, up to 60s) makes this a
   low-probability case, and the failure mode is a graceful `console.warn`
   + empty comments feed rather than a crashed build, so this is a
   reasonable trade for the simplicity gained — just worth the author (or a
   future reader) knowing the blast radius changed from "one issue" to "all
   comments, this run."

2. **The 5,000-comment cap is a real trade-off, honestly disclosed but only
   in the code comment, not the PR body's headline claims.** The in-code
   comment says it plainly ("the oldest closed threads thin out gracefully")
   and the PR body's "Growth flattened & bounded" bullet gestures at it, but
   a reader skimming just the PR description could come away thinking this
   is a pure win with no data trade-off. It isn't hidden, and at the repo's
   current size (30 of a possible 50 pages) it changes nothing today — just
   flagging that once total repo comments pass ~5,000, the oldest comments
   on old, low-traffic issues will start silently aging out of that issue's
   `commentsList`, which the old code never did (it fetched a given issue's
   comments in full, regardless of repo-wide age, as long as that issue had
   ≤60 comments). Not a blocker; the cap is a sane, bounded-cost design
   choice for a feed that's inherently "recent activity" focused.

## Fit / scope / safety

- Fit: follows the file's existing conventions (reuses `ghPaged`, same
  header/auth/retry plumbing, same map-building style as the rest of the
  script).
- Scope: one file, no new infrastructure, no governance/ADR/label/pipeline
  changes — doesn't trigger the human-ratification guard.
- Safety: no secrets, no personal/identifying data. No behavior affecting
  the research pipeline, labels, or review gates.
- Honesty of framing: the PR body's numbers (557→~12 calls, 191s→45s,
  parity on #307/#36/#16/a PR) are verifiable claims, not decisions
  presented as already-ratified; I independently reproduced the #307 parity
  and the call-count/latency order of magnitude.

## Verdict

Correct, verified by an actual run against the live API with matching
output, self-contained, in scope, no safety issues. The two notes above are
worth a maintainer's awareness but don't block this PR.

VERDICT: PASS

